PR #26, opened 2018-03-20
Enable automated npm publishing linked with GitHub releases

### LINKED ISSUE #24: Create automated releases
Might be interesting for us:
https://hackernoon.com/automate-npm-releases-with-semantic-release-and-human-written-change-logs-2adb1dce487

Acceptance tests:
- [ ] npm package is automatically published after merge when changing version in package.json.

jaroslav-kubicek:
Closes #24 

Idea:

- Publishing is tightly coupled with GitHub releases https://github.com/kiwicom/smart-faq/releases
- Each time we create a new release, npm package is automatically published

As a result, we will gain clean changelog with descriptions of all changes.

I decided not to do `npm version patch` for each commit - it's not feasible when we are working simultaneously and it would lead to a large number of releases, but it might be possible after merging into master 🤔

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,13 +3,13 @@
 # Check https://circleci.com/docs/2.0/language-javascript/ for more details
 
 version: 2
+defaults: &defaults
+  working_directory: ~/smart-faq
+  docker:
+    - image: circleci/node:8.0
 jobs:
-  build: # smart-faq
-    docker:
-      - image: circleci/node:8
-
-    working_directory: ~/smart-faq
-
+  build:
+    <<: *defaults
     steps:
       - checkout
 
@@ -30,8 +30,7 @@ jobs:
       - run: yarn test-ci
 
   deploy:
-    docker:
-      - image: circleci/node:8
+    <<: *defaults
     steps:
       - checkout
       # Download and cache dependencies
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -42,6 +41,23 @@ jobs:
...
+      - checkout
+      - run:
+          name: Write NPM Token to ~/.npmrc
+          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
+      - run:
+          name: Install dot-json package
+          command: npm install dot-json
+      - run:
+          name: Write version to package.json
+          command: $(yarn bin)/dot-json package.json version ${CIRCLE_TAG:1}
```
git-toni: Is this actually being written in the `package.json` found in the repo?

jaroslav-kubicek: yes, and then it's thrown away, version in package.json isn't really needed except when running npm publish

jaroslav-kubicek: if we don't like it, then we can switch to https://github.com/semantic-release/semantic-release or just trigger it by changing version in package.json, but that would leave us without proper changelog on the other hand

jaroslav-kubicek: or I can just commit changes in package.json after publish of course 🙂

git-toni: I think writing changes after publish would be the best approach :)

jaroslav-kubicek: should be updated

PR #41, opened 2018-03-24
12 solve login

### LINKED ISSUE #12: Solve how we will deal with KIWICOM login
Find a way how we're going to handle log in via Kiwi, Facebook & Google in cooperation with frontend

### LINKED ISSUE #16: State management
Use cases that motivate some sort of pre-planned state management:
- Have access to current language. Eg. more than one component will need to make requests to API passing language as a query/header param.
- If locales(aka the actual translation contents) are finally passed via props, any component should be able to access them.
- When transitioning from one screen to the other, the current state should be kept, eg. the search query, the FAQ depth level, input boxes text...

Options so far:
- Centralized store(MobX, Redux, etc)
- Component state -based approach
- Other? Singletons?

git-toni:
Closes #12, closes #16 

Notes:
- Use of Redux to store Token.
- Login status indicator at `Layout` component, out of the main screen, temporary for debugging.
- Added `dotenv-webpack` to load an env_variable required to Login to kiwicom. I assumed this was sensible data so better have as a variable in `.env`, particularly `KIWILOGIN_USER` (value passed via Slack).
- This makes use of Token saved at redux store + cookie. Other login info used at kiwicom seems irrelevant for now for smart-faq.

Todo:
- [X] Mocks and tests


src/HOC/withAuth.js
```diff
@@ -0,0 +1,40 @@
...
+      return <BaseComponent {...this.props} />;
+    }
+  }
+  return connect(mapStateToProps, mapDispatchToProps)(WithAuth);
+}
+
+const mapStateToProps = (state: StoreType): SessionState => ({
+  ...state.session,
+});
+const mapDispatchToProps = (dispatch: Function) => ({
```
jaroslav-kubicek: could be defined just as an object

```
const mapDispatchToProps = {
  doSignIn: signIn,
  doSaveToken: saveToken
}
```

src/HOC/withAuth.js
```diff
@@ -0,0 +1,40 @@
...
+};
+
+function withAuth(BaseComponent: React.ComponentType<any>) {
+  class WithAuth extends React.Component<Props> {
+    constructor(props) {
+      const JWT_Token = getCookieToken();
+      if (JWT_Token) {
+        props.doSaveToken(JWT_Token);
+      }
+      super(props);
```
jaroslav-kubicek: I would personally advice to call `super` as the first thing in constructor, although it's not necessary until you use `this`

src/HOC/withAuth.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+import * as React from 'react';
+import { connect } from 'react-redux';
+import type { StoreType } from '../store/reducers';
+import type { State as SessionState } from '../store/reducers/SessionReducer';
+import { saveToken, signIn } from '../store/actions/SessionActions';
+import { getCookieToken } from '../helpers/Auth';
+
+type Props = {
+  doSignIn: Function,
```
jaroslav-kubicek: I would be more specific about the function types => e.g. `doSaveToken: (token: string) => void`

webpack.config.js
```diff
@@ -4,12 +4,13 @@
 const path = require('path');
 const webpack = require('webpack');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
+const Dotenv = require('dotenv-webpack');
```
jaroslav-kubicek: README file should be updated with necessary information

src/helpers/Auth.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+import Cookies from 'js-cookie';
+import { Requester } from './Requests';
+
+export const COOKIE_LOGIN_KEY = 'ua_session_token';
```
jaroslav-kubicek: I would prefer more specific/obvious key, e.g. `faq_ua_session_token`

git-toni: Are you referring to the value of the cookie key? ie.`ua_session_token`.

If so, I think we should keep it as that since this is the one kiwicom uses and otherwise we wouldn't share the same session with them.

jaroslav-kubicek: oh, I see, ok then

PR #51, opened 2018-03-27
Add Apollo client provider to enable GraphQL queries

jaroslav-kubicek:
I'm kicking off GraphQL with [apollo-boost](https://github.com/apollographql/apollo-client/tree/master/packages/apollo-boost) which already contains all that is necessary for the beginning.


src/App.js
```diff
@@ -2,14 +2,20 @@
...
 import initTranslation from './initTranslation';
+import store from './store';
 import Routes from './Routes';
+import Layout from './Layout';
 
 type Props = {
   locale: Object,
 };
 
+const client = new ApolloClient({ uri: process.env.GRAPHQL_URI });
```
git-toni: Should we maybe add the required `Auth` fields here already? Some end-points require it.

We can also leave this for another issue, and plan it a bit more scalably...

image

jaroslav-kubicek: We shouldn't need it for FAQ queries and if so, we can always add it later

jaroslav-kubicek: ha, you might need it for bookings - then I would do it in PR related to that 🙂

git-toni: Yep, I tried bookings first thing, so that's why it failed. But I agree it's better to move it to another PR, and better plan it. Since we will encounter the case where we initialize the `client` without token, and then, once the user logs-in, will need to re-configure the client with the new token. :)

PR #55, opened 2018-03-28
Migrate to relay modern instead of Apollo client

jaroslav-kubicek:


package.json
```diff
@@ -73,21 +79,21 @@
   },
   "dependencies": {
     "@kiwicom/orbit-components": "^0.0.0-rc1",
-    "apollo-boost": "0.1.4",
     "axios": "^0.18.0",
-    "graphql": "0.13.2",
     "babel-runtime": "^6.26.0",
     "i18next": "10.5.0",
+    "isomorphic-fetch": "2.2.1",
```
git-toni: Wouldn't `axios` be enough?

git-toni:
I think this PR should include the `get-graphql-schema` package.json script to retrieve the updated version of the schema from the graphql repo endpoint.

jaroslav-kubicek:
I added `graphql-cli` instead for the same purpose, which should be better choice, cc @git-toni

package.json
jaroslav-kubicek: `axios` is actually adding additional functionality which we don't need - transformers, interception or cancellation

git-toni:
@jaroslav-kubicek  Yep, sorry, missed the `.graphqlconfig` settings. Much better like this I agree :)

package.json
git-toni: Agree, then we should replace all `axios` uses with polyfilled `fetch`. I'll raise  an issue for that

Issue #11, opened 2018-03-06
Add navigation through static FAQ categories


jaroslav-kubicek:
Acceptance criteria:
- User can navigate through subcategories of FAQs
- User can jump back to parent category by clicking on "back"
- User can navigate through parent categories by clicking on breadcrumbs

**note**
- transition between categories should be prepared by design team
- ~we don't have API for this(related to login status)~

Design: https://marvelapp.com/7iihb90/screen/39265582

MagdalenaAlmanova:
I found one thing. There was the` "Back"` button missing.
![screen shot 2018-04-26 at 21 20 39](https://user-images.githubusercontent.com/37298895/39327484-6fdc1550-4998-11e8-8850-c18acf760a70.png)

_Note: The text of some subcategories was out of the box._
![screen shot 2018-04-26 at 21 22 00](https://user-images.githubusercontent.com/37298895/39327529-92d8947a-4998-11e8-9998-b80274cc9fc8.png)

Thanks,
Meggy

MagdalenaAlmanova:
Two mistakes have been fixed, but I am not sure if correctly. In case the question is longer, shouldn't it continue on the next line?
**Demo v.**
![screen shot 2018-05-10 at 00 11 18](https://user-images.githubusercontent.com/37298895/39842465-c019cbae-53e6-11e8-95f9-017e1dae24cf.png)

**Prototype v.**
![screen shot 2018-05-10 at 00 12 30](https://user-images.githubusercontent.com/37298895/39842500-ddedb348-53e6-11e8-97ea-9a0ead28eeb2.png)

Thank you for your answer :)
M.

MagdalenaAlmanova:
It was my mistake, I have forgotten to assign you, sorry.

I checked it again and the recognized mistake was repaired. I found also something else. The description of the question doesn't continue to the next line as in the zeplin 1b-2a-1. 
Demo v.
![screen shot 2018-05-15 at 21 08 12](https://user-images.githubusercontent.com/37298895/40078191-3ab206f0-5884-11e8-9187-eef3aa1c3a95.png)

Prototype v.
![screen shot 2018-05-15 at 21 03 22](https://user-images.githubusercontent.com/37298895/40078134-0dc26a04-5884-11e8-8b82-e226fd30e374.png)
 

Thanks :) @jalbertsr

jalbertsr:
[...]

PR #62, opened 2018-04-04
Add navigation through FAQ subcategories

jaroslav-kubicek:
References #11 

- it's possible to click on one of the top level categories
- it also renders perexes of FAQ articles - not clickable now
- no breadcrumbs for now<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-nosadwwvyq.now.sh
</url>

### DIFF
--- a/src/StaticFAQ/FAQCategoryList.js
+++ b/src/StaticFAQ/FAQCategoryList.js
@@ -2,21 +2,35 @@
 
 import * as React from 'react';
 import idx from 'idx';
+import { Link } from 'react-router-dom';
 import { graphql, QueryRenderer } from 'react-relay';
 
+import FAQArticle from './FAQArticle';
 import FAQCategory from './FAQCategory';
 import environment from '../relay/environment';
-import type { FAQCategoryListQueryResponse } from './__generated__/FAQCategoryListQuery.graphql';
+import routeDefinitions from './../routeDefinitions';
 
-type Props = {};
+import type { FAQArticle_article } from './__generated__/FAQArticle_article.graphql';
+import type { FAQCategory_category } from './__generated__/FAQCategory_category.graphql';
+import type { FAQCategoryListRootQueryResponse } from './__generated__/FAQCategoryListRootQuery.graphql';
+import type { FAQCategoryListSubcategoryQueryResponse } from './__generated__/FAQCategoryListSubcategoryQuery.graphql';
 
-type QueryRendererParams = {
-  props: FAQCategoryListQueryResponse,
+type Props = {|
+  categoryId: string | null,
+|};
+
+type RootQueryRendererParams = {
+  props: FAQCategoryListRootQueryResponse,
   error: Error,
 };
 
-const query = graphql`
-  query FAQCategoryListQuery {
+type SubcategoryQueryRendererParams = {
+  props: FAQCategoryListSubcategoryQueryResponse,
+  error: Error,
+};
+
+const queryRoot = graphql`
+  query FAQCategoryListRootQuery {
     allFAQCategories(language: en) {
       edges {
         node {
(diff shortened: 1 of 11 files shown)


src/StaticFAQ/FAQCategoryList.js
```diff
@@ -2,66 +2,138 @@
...
+        id
+        ...FAQItemArticle
+      }
+    }
+  }
+`;
 
 class FAQCategoryList extends React.Component<Props> {
-  renderStaticFAQs = (rendererProps: QueryRendererParams) => {
+  renderFAQArticlePerexes = (faqs: Object[]) => {
```
git-toni: We can probably 
```
import type { FAQItemArticle}  from './__generated__/FAQItemArticle.graphql';
```
And use that type instead of `Object` here.

src/StaticFAQ/FAQCategoryList.js
```diff
@@ -2,66 +2,138 @@
...
+    return <div>{faqs.map(faq => <FAQArticle key={faq.id} data={faq} />)}</div>;
+  };
+
+  renderCategories = (categories: Object[]) => {
```
git-toni: We can probably 
```
import type { FAQItemCategory}  from './__generated__/FAQItemCategory.graphql';
```
And use that type instead of `Object` here.

src/StaticFAQ/FAQCategoryList.js
```diff
@@ -2,66 +2,138 @@
...
+  };
+
+  renderRootCategory = (rendererProps: RootQueryRendererParams) => {
     if (rendererProps.error) {
       return <div>Error</div>;
     }
 
     if (rendererProps.props) {
-      const categories =
+      const edges =
```
git-toni: What about moving these two lines(get edges, and then get nodes) to  general function? We can also wait to see how many times we'll repeat before DRYing as well.

git-toni:
This has some great ideas :+1: 

A couple of design problems I noticed are:
- Overflow of rows at the FAQ. Still design didn't cover this, maybe with a scroll? So we can merge this while that's not designed.
- Articles' perex is overflowing their container. That is covered by design now.

Eg.
Cagetories overflow:
image


Article perex overflow:
image

src/StaticFAQ/FAQItem.js
```diff
@@ -31,34 +29,48 @@ const styles = css`
...
+  graphql`
+    fragment FAQItemCategory on FAQCategory {
+      id
+      title
+    }
+  `,
+);
+
+export const FAQArticle = createFragmentContainer(
+  FAQItem,
```
jalbertsr: Just notice that article and categories have a slightly different design, I'll be creating a common article card/box component, so for now is ok.

jalbertsr:
@git-toni about your https://github.com/kiwicom/smart-faq/pull/62#issuecomment-379206641:
- Overflow of rows at the FAQ is covered in zeplin, and has to be resolved by scrolling. 
- Also this PR, I guess is not closing the issue, just doing the issue logic.

git-toni:
@jalbertsr Agree, comment deleted :)

src/StaticFAQ/FAQItem.js
jaroslav-kubicek: Ha! ok, I see, then I will refactor that into two independent components

src/StaticFAQ/FAQCategoryList.js
jaroslav-kubicek: good point, fixed

src/StaticFAQ/FAQCategoryList.js
jaroslav-kubicek: I would wait since it's not needed now 🙂

jaroslav-kubicek:
Ok, I updated this PR, I created `Card` component as well, yet it's slightly different - it's just dumb wrapper adding some styles around FAQ or category

PR #72, opened 2018-04-10
Adjust smart FAQ to be used as dependency

jaroslav-kubicek:


package.json
```diff
@@ -4,7 +4,7 @@
   "scripts": {
     "preinstall": "npm run build",
     "start": "NODE_ENV=development LANGUAGE=en webpack-dev-server --open --mode development",
-    "build": "npm run clean && npm run build:lib && npm run build:flow",
+    "build": "npm run clean && npm run build:lib",
```
jaroslav-kubicek: just to clarify - raw flow files wouldn't work now on frontend

PR #76, opened 2018-04-11
update query fields

jalbertsr:


src/StaticFAQ/SearchAllFAQs.js
```diff
@@ -90,20 +90,16 @@ class SearchAllFAQs extends React.Component<Props> {
...
+  renderFAQs = (faqList: FAQs[]) => (
     <div className="scrollable-box">
-      {results.map(result => (
-        <FAQArticle
-          key={result.articleId}
-          title={result.title}
-          perex="How to find your best deal fast and easy"
-        />
+      {faqList.map(faq => (
+        <FAQArticle key={faq.id} title={faq.title} perex={faq.perex} />
```
jaroslav-kubicek: this is not sufficient change, please use `FAQArticle` with relay fragments & remove redundant `FAQArticle` in this file

PR #101, opened 2018-04-19
69 - Add StaticFAQ cache

### LINKED ISSUE #69: Add caching for static FAQs
We should probably introduce some sort of caching at least for static FAQs & categories, to avoid repetitive requests for the exactly same data over and over again and make navigation thought FAQs faster.

git-toni:
Closes #69 

Caching only FAQ-related routes. Booking-related ones are better to re-fetch for now, due to different component renders depending on time left.


src/relay/environment.js
```diff
@@ -1,15 +1,30 @@
...
+  QueryResponseCache,
+} from 'relay-runtime';
 import { getSessionToken } from '../helpers/Auth';
 
 require('isomorphic-fetch');
 // used when smart FAQ installed as dependency
 const uri = 'https://graphql.kiwi.com';
+const cache = new QueryResponseCache({ size: 200, ttl: 30 * 60 * 1000 });
+
+const shouldRouteCache = (queryName: string) => queryName.startsWith('FAQ');
```
jaroslav-kubicek: this isn't too scalable, instead, `fetchQuery` can accept the third param, `cacheConfig` with property `force`. IMHO It should be decided upon a value of this prop whether we should use cache or not 🙂 

... and then pass `cacheConfig=({force: true}}` to every `<QueryRenderer>` that should always fetch fresh data

src/relay/environment.js
```diff
@@ -20,7 +35,11 @@ const buildFetchQuery = (token: string = '') => {
         query: operation.text, // GraphQL text from input
         variables,
       }),
-    }).then(response => response.json());
+    }).then(response => {
+      const data = response.json();
```
jaroslav-kubicek: `response.json()` returns promise AFAIK: https://developer.mozilla.org/en-US/docs/Web/API/Body/json

```
.then(response => response.json())
.then(data => { 
  routeCached && cache.set(operation.text, variables, data);
  return data;
})
```

src/relay/environment.js
git-toni: Absolutely agree, this was a first version of the route "whitelisting" but your suggestions sounds definitely better, thx.

PR #110, opened 2018-04-23
Refactor login to make it work with frontend

jaroslav-kubicek:
🎉 Login integration with Frontend is coming!

LiveURL: ~https://kiwicomsmart-faq-yhzjdtoioz.now.sh~ 
https://kiwicomsmart-faq-paperwmyfe.now.sh


src/common/index.js
```diff
@@ -1,7 +0,0 @@
-// @flow
```
git-toni: I think this file should be kept, and its use more encouraged, since we do not use it a lot currently :slightly_smiling_face:

git-toni:
Great job! And suitable way to integrate fully with frontend :) One suggestion, one question: I 

- Sug: All the context-related things you add in `src/common` could be instead added to a location such as `src/context` or `src/store`, and keep the `src/common` for recurrent UI-related components.
- Q: Do the functions `onLogin`, `onLogout`, etc that the new `src/App` is expecting exist in the frontend codebase as one? I mean eg. one function that integrates all the login logic(write cookies, write their redux store...), and another one that integrates all the logout logic.

jaroslav-kubicek:
> Do the functions onLogin, onLogout, etc that the new src/App is expecting exist in the frontend codebase as one? I mean eg. one function that integrates all the login logic(write cookies, write their redux store...), and another one that integrates all the logout logic.

Check out our PR on frontend, you can also try it yourself on local machine - just setup frontend locally  and then something like `yarn build` in smart FAQ repo, and in frontend repo: `rm -rf ./node_modules`, `yarn add ./path/to/smartFAQ` (you need to remove smart faq from package.json probably before) and `yarn install`

src/index.js
```diff
@@ -1,10 +1,80 @@
...
+class Root extends React.Component<Props, State> {
+  cookieKey: string;
+
+  constructor(props) {
+    super(props);
+
+    this.cookieKey = 'mockedLogin';
+
+    this.state = {
+      user: null,
```
git-toni: In order to be able to keep the session after refresh when testing on github pages what if we:
1. Move the `user` in Line 37 out of the class.
2. Then replace this block in the constructor with:
```
    const loginToken = Cookies.get(this.cookieKey);
    this.state = {
      user: loginToken ? user : null,
      loginToken,
    };
```

src/index.js
```diff
@@ -1,10 +1,81 @@
...
+          language="en"
+          locale={enLocale}
+          user={this.state.user}
+          loginToken={this.state.loginToken}
+        />
+        <style jsx global>
+          {`
+            .root {
+              position: absolute;
+              right: 0;
```
jalbertsr: I see a double scrollbar 

screnshot

probably adding `top: 0` resolves it

jaroslav-kubicek: fixed, thx

jaroslav-kubicek:
Working url: https://kiwicomsmart-faq-ymjzxiotlg.now.sh

PR #117, opened 2018-04-25
add back arrow

### LINKED ISSUE #109: Add Back arrow to breadcrumbs
Some designs show a back arrow button to move backwards along the breadcrumbs. https://app.zeplin.io/project/5a8e8992554d385c76ad5aeb/screen/5aaac0e38d835c7b25745c15

It seems only the logged-in FAQ shows that arrow, whereas the logged-out case does not. We should clarify that and add it to the app.
image

jalbertsr:
Close #109 <br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-uyvwrmnisg.now.sh
</url>

### DIFF
--- a/src/StaticFAQ/Breadcrumbs.js
+++ b/src/StaticFAQ/Breadcrumbs.js
@@ -1,9 +1,12 @@
 // @flow
 
 import * as React from 'react';
+import { Link } from 'react-router-dom';
 import { createFragmentContainer, graphql } from 'react-relay';
+import { ChevronLeft } from '@kiwicom/orbit-components/lib/icons';
 
 import Breadcrumb from './Breadcrumb';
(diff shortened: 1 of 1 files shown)


src/StaticFAQ/Breadcrumbs.js
```diff
@@ -1,19 +1,63 @@
...
 type Props = {|
   breadcrumbs: Breadcrumbs_breadcrumbs,
   currentCategory: string,
 |};
 
+type BackArrowProp = {|
+  id: string,
+|};
+
+const BackArrow = ({ id }: BackArrowProp) => {
```
jaroslav-kubicek: I would also define this in a separate file 🙂  👼

Issue #29, opened 2018-03-21
Search in static FAQs


jaroslav-kubicek:
**Acceptance criteria:**
- when user clicks on "continue without booking", he can see list of FAQ categories and input above them
- user can start typing and see results matching the input text
- each item in the list of results is FAQ article preview (differs from category by ">" on right) and no category is shown in the list
- when user clears the input, list of original FAQ categories is shown
- screen looks just like in the design
- clicking on "x" in input clears the input and list of top-level FAQs categories is shown

<img width="383" alt="screen shot 2018-03-21 at 11 23 05" src="https://user-images.githubusercontent.com/2174629/37707279-8b5fd060-2cfa-11e8-8dab-3c2034da7770.png">

https://marvelapp.com/7iihb90/screen/39740344

MagdalenaAlmanova:
I found only one defect: When I clicked on (x) the input text was not deleted. It was only on Firefox 59.0.2 (64bit), MacOs High Sierra 10.13.4.

I have a suggestion as an user: The scrollbar should be permanently visible in order to natively indicate that there are more articles to follow in below.

Thank you,
Meggy

MagdalenaAlmanova:
Perfect job but I think I  have found something. When I typed for example "baggage or price" and It showed me results, there are some articles where text was out of the box. This does not look good.  As you can see on print screens:

![screen shot 2018-04-11 at 20 34 49](https://user-images.githubusercontent.com/37298895/38636416-69d94014-3dc8-11e8-83bd-17bfb3b65d06.png)

![screen shot 2018-04-11 at 20 34 01](https://user-images.githubusercontent.com/37298895/38636429-6f50087a-3dc8-11e8-9a68-7be3ce5f7082.png)

Tested on **MacOs High Sierra 10.13.4**. browsers: _Safari, Chrome, Opera, Firefox last versions_.

Thank you,

MRTNKCR:
This will probably require some changes in Knowledgebase, we already have a spec for that, please check it out here: https://paper.dropbox.com/doc/Knowledgebase-additional-features-SMart-FAQ-1.0.-JB8W3H8ptSG5LCvb7jtuJ

jalbertsr:
I would like to close this issue since search faq main logic works, I will create a new issue with the pending problems.

MagdalenaAlmanova:
Yes, good idea.

PR #125, opened 2018-04-27
Fix <Card> size when the header is too big

### LINKED ISSUE #87: faq articles in search have too large headings.
Text is out of the box for FAQ articles in search.
Please see discussion on #29 for more details.

Match design: [Zeplin](https://app.zeplin.io/project/5a8e8992554d385c76ad5aeb/screen/5aa98f807e4ac0f30f266e1c)

kantord:
fixes #87 <br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-cwkfpnwtvq.now.sh
</url>
</url>

### DIFF
--- a/src/common/Card.js
+++ b/src/common/Card.js
@@ -8,7 +8,7 @@ const styles = css`
     padding: 14px 22px;
     position: relative;
     margin-top: 24px;
-    height: 100px;
+    min-height: 100px;
     width: 100%;
     border-radius: 3px;
     box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.12);
--- a/src/common/Card.js
+++ b/src/common/Card.js
@@ -8,7 +8,7 @@ const styles = css`
     padding: 14px 22px;
     position: relative;
     margin-top: 24px;
-    min-height: 100px;
+    height: 100px;
     width: 100%;
     border-radius: 3px;
     box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.12);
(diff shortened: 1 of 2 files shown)


src/common/Card.js
```diff
@@ -8,7 +8,7 @@ const styles = css`
     padding: 14px 22px;
     position: relative;
     margin-top: 24px;
-    height: 100px;
+    min-height: 100px;
```
jaroslav-kubicek: it was lost in discussion in #29  but instead of flexible height, the text should be cut with `...` at the end so the card has always the same height

according to this doc:
 
 https://paper.dropbox.com/doc/Knowledgebase-additional-features-Smart-FAQ-1.0.-JB8W3H8ptSG5LCvb7jtuJ

these should be length limits:
```
​Categories:
​​Headings: 45 characters
Perex: 90 characters
​​
​​Articles:
​​Headings: 90 characters
Perex: 100 characters
```

kantord: hmm, ok, I can do that. What do you think about using a CSS ellipsis overflow instead, though? That way it's possible to define the maximum length (and/or) with CSS and have an '...' if it's too long. It seems more stable to me, as it can maximize the number of characters that can be displayed, without the possibility of an overflow. Limiting at a fixed number of characters might still cause an overflow if the characters are unusually wide, or else, we might up wasting screen space (cutting of characters when there's plenty of space left)

kantord: I undid the height thing, and implemented the header thing the `text-overflow: ellipsis` way.

PR #126, opened 2018-04-30
Add <Breadcrumbs /> to FAQArticleDetailContent

kantord:
<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-bnqpdmufci.now.sh
</url>
</url>
</url>

### DIFF
--- a/src/ContentPage/index.js
+++ b/src/ContentPage/index.js
@@ -53,7 +53,7 @@ const FAQRoutes = (
       />
       <Route
         exact
-        path={`${routeDefinitions.FAQ_ARTICLE}/:articleId`}
+        path={`${routeDefinitions.FAQ_ARTICLE}/:categoryId/:articleId`}
         component={FAQArticleDetail}
       />
     </Switch>
--- a/src/StaticFAQ/Breadcrumbs.js
+++ b/src/StaticFAQ/Breadcrumbs.js
@@ -2,10 +2,8 @@
 
 import * as React from 'react';
 import { createFragmentContainer, graphql } from 'react-relay';
-import css from 'styled-jsx/css';
 
 import CustomBreadcrumbs from './CustomBreadcrumbs';
-import BackArrow from './BackArrow';
 import type { Breadcrumbs_breadcrumbs } from './__generated__/Breadcrumbs_breadcrumbs.graphql';
 
 type Props = {|
@@ -13,10 +11,14 @@ type Props = {|
   currentCategory: string,
 |};
 
-const Breadcrumbs = ({ breadcrumbs, currentCategory, loginToken }: Props) => {
-  return (<CustomBreadcrumbs breadcrumbs={
-    [{title: 'Home'}].concat(breadcrumbs).concat([{title: currentCategory}])
-  } />)
+const Breadcrumbs = ({ breadcrumbs, currentCategory }: Props) => {
+  return (
+    <CustomBreadcrumbs
+      breadcrumbs={[{ title: 'Home' }]
+        .concat(breadcrumbs)
+        .concat([{ title: currentCategory }])}
+    />
+  );
 };
 
 export default createFragmentContainer(
--- a/src/StaticFAQ/Breadcrumbs.js
+++ b/src/StaticFAQ/Breadcrumbs.js
@@ -2,11 +2,20 @@
 
 import * as React from 'react';
 import { createFragmentContainer, graphql } from 'react-relay';
+import css from 'styled-jsx/css';
 
 import Breadcrumb from './Breadcrumb';
 import BackArrow from './BackArrow';
 import { withLoginToken } from '../context/User';
 import type { Breadcrumbs_breadcrumbs } from './__generated__/Breadcrumbs_breadcrumbs.graphql';
+
+const style = css`
+  .breadcrumbs {
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
+  }
+`;
 
 type Props = {|
   breadcrumbs: Breadcrumbs_breadcrumbs,
@@ -20,21 +29,25 @@ const Breadcrumbs = ({ breadcrumbs, currentCategory, loginToken }: Props) => {
 
   return (
     <React.Fragment>
-      {loginToken && <BackArrow id={id} />}
-      <Breadcrumb breadcrumb={{ title: 'Home' }} />
-      {breadcrumbs.map(breadcrumb => {
-        if (breadcrumb && breadcrumb.title) {
-          return (
-            <Breadcrumb
-              key={breadcrumb.id}
-              breadcrumb={{ id: breadcrumb.id, title: breadcrumb.title }}
-            />
-          );
-        }
+      <div className="breadcrumbs">
+        {loginToken && <BackArrow id={id} />}
+        <Breadcrumb breadcrumb={{ title: 'Home' }} />
+        {breadcrumbs.map(breadcrumb => {
+          if (breadcrumb && breadcrumb.title) {
+            return (
+              <Breadcrumb
+                key={breadcrumb.id}
+                breadcrumb={{ id: breadcrumb.id, title: breadcrumb.title }}
+              />
+            );
+          }
 
-        return null;
-      })}
-      <Breadcrumb breadcrumb={{ title: currentCategory }} isCurrent />
+          return null;
+        })}
+        <Breadcrumb breadcrumb={{ title: currentCategory }} isCurrent />
+      </div>
+
+      <style jsx>{style}</style>
     </React.Fragment>
   );
 };
--- a/src/StaticFAQ/Breadcrumbs.js
+++ b/src/StaticFAQ/Breadcrumbs.js
@@ -28,27 +28,24 @@ const Breadcrumbs = ({ breadcrumbs, currentCategory, loginToken }: Props) => {
   const id = lastCategory && lastCategory.id;
 
   return (
-    <React.Fragment>
-      <div className="breadcrumbs">
-        {loginToken && <BackArrow id={id} />}
-        <Breadcrumb breadcrumb={{ title: 'Home' }} />
-        {breadcrumbs.map(breadcrumb => {
-          if (breadcrumb && breadcrumb.title) {
-            return (
-              <Breadcrumb
-                key={breadcrumb.id}
-                breadcrumb={{ id: breadcrumb.id, title: breadcrumb.title }}
-              />
-            );
-          }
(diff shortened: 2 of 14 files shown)


src/StaticFAQ/FAQArticleDetail.js
```diff
@@ -37,14 +52,23 @@ class FAQArticleDetail extends React.Component<Props> {
...
+        />
+      );
     }
 
     return <Loader />;
   };
 
   render() {
     const articleId = idx(this.props.match, _ => _.params.articleId);
+    const categoryId = this.props.history.entries
```
kantord: Although it should always work, I'm not entirely satisfied with this part. Do you have a suggestion that might improve it? :-)

src/StaticFAQ/FAQArticleDetail.js
```diff
@@ -11,10 +11,22 @@ import { LanguageContext } from '../context/Language';
...
+  ) {
     FAQArticle(id: $id, language: $language) {
       ...FAQArticleDetailContent_article
     }
+    FAQCategory(id: $category_id) {
+      id
+      title
+      ancestors {
+        id
+        ...Breadcrumbs_breadcrumbs
```
kantord: Here the current category should be included as well. As we've discussed, I had not implemented it yet, but should it be implemented now? Do you have a suggestion?

src/StaticFAQ/Breadcrumbs.js
```diff
@@ -20,21 +29,25 @@ const Breadcrumbs = ({ breadcrumbs, currentCategory, loginToken }: Props) => {
 
   return (
     <React.Fragment>
```
jaroslav-kubicek: then I guess `React.Fragment` is not needed anymore

src/StaticFAQ/FAQArticleDetail.js
jaroslav-kubicek: I would definitely suggest changing the url so it includes `categoryId` , e.g. `/categories/{categoryId}/articles/{articleid}`

src/StaticFAQ/FAQArticleDetailContent.js
```diff
@@ -5,58 +5,65 @@ import Markdown from 'markdown-to-jsx';
...
 import { createFragmentContainer, graphql } from 'react-relay';
 
-import type { FAQArticleDetailContent_article } from './__generated__/FAQArticleDetailContent_article.graphql';
+import Breadcrumbs from './Breadcrumbs';
+import type { FAQArticleDetailQuery } from './__generated__/FAQArticleDetailQuery.graphql';
 
-type Props = {
-  article: FAQArticleDetailContent_article,
-};
+type Props = FAQArticleDetailQuery;
```
jaroslav-kubicek: this looks suspicious, is there `FAQArticleDetailQuery` type?

old `article: FAQArticleDetailContent_article` should be kept, that's why it was generated

src/StaticFAQ/FAQArticleDetail.js
```diff
@@ -11,10 +11,22 @@ import { LanguageContext } from '../context/Language';
...
+  query FAQArticleDetailQuery(
+    $id: ID!
+    $language: Language
+    $category_id: ID!
+  ) {
     FAQArticle(id: $id, language: $language) {
       ...FAQArticleDetailContent_article
     }
+    FAQCategory(id: $category_id) {
+      id
```
jaroslav-kubicek: it fetches `id`, `title` and `ancestors.id` for this component but it's never used

src/StaticFAQ/FAQArticleDetailContent.js
```diff
@@ -5,58 +5,65 @@ import Markdown from 'markdown-to-jsx';
...
-      {article.title}
-    </Typography>
-    <div className="faq-article-perex">
-      <Typography type="attention" variant="bold">
-        Summary:
+const Detail = ({ article, category }: Props) => (
+  <React.Fragment>
+    <div className="faq-article-content">
+      <div>
+        <Breadcrumbs
```
jaroslav-kubicek: Breadcrumbs don't need to be in this component, moving them to `renderDetailContent` might help you to define relay fragments in a better way

kantord:
@jaroslav-kubicek I have made the requested changes + resolved merge conflicts.

However, I get flow errors on the merge commit, and I don't understand why. I'm a little stuck. Do you have a suggestion? :-)

src/StaticFAQ/FAQCategoryList.js
```diff
@@ -72,12 +73,13 @@ const querySubcategory = graphql`
...
   renderFAQArticlePerexes = (
     faqs: $ReadOnlyArray<?FAQArticlePerexFragment>,
+    category: FAQCategory_category,
   ) => {
     return (
       <div>
         {faqs
           .filter(Boolean)
-          .map(faq => <FAQArticle key={faq.id} article={faq} />)}
+          .map(faq => <FAQArticle key={faq.id} article={faq} category={category} />)}
```
jalbertsr: if you add category to `FAQArticle`, you have to add it also to the `FAQArticle` on `SearchAllFAQs.js`.

git-toni:
@kantord I have a suggestion for another approach. 
If you decorate the export of `src/StaticFAQ/FAQArticle.js` with the `withRouter` provided by react-router, you will have access to all the past entries of history - particularly at `props.history.entries` of that component.

I.e.
```
//Rest of src/StaticFAQ/FAQArticle.js
export default createFragmentContainer(
  withRouter(FAQArticle),
  graphql`
    fragment FAQArticle_article on FAQArticle {
      id
      title
      perex
    }
  `,
);
```
Wouldn't this solve the Breadcrumbs info-requirements?

image

jalbertsr:
notice breadcrumb designs for articles that come from search or articles that are found from going through the tree structure are different: 

sreenshot

screenshot

src/StaticFAQ/FAQArticle.js
```diff
@@ -9,9 +9,11 @@ import css from 'styled-jsx/css';
 import Card from './../common/Card';
 import routeDefinitions from '../routeDefinitions';
 import type { FAQArticle_article } from './__generated__/FAQArticle_article.graphql';
+import type { FAQCategory_category } from './__generated__/FAQCategory_category.graphql';
 
 type Props = {|
   article: FAQArticle_article,
+  category: FAQCategory_category,
```
jaroslav-kubicek: passing just `categoryId` string is enough

src/ContentPage/index.js
```diff
@@ -53,7 +53,7 @@ const FAQRoutes = (
       />
       <Route
         exact
-        path={`${routeDefinitions.FAQ_ARTICLE}/:articleId`}
+        path={`${routeDefinitions.FAQ_ARTICLE}/:categoryId/:articleId`}
```
git-toni: I'm not sure this is a good idea. 

Articles can belong to several categories, so it's conceptually a bit confusing IMHO. Also some articles will be reached after more than one nested category screen and this route does not cover those cases.

What do you think? :slightly_smiling_face:

kantord: The idea is the each article would have multiple URLs instead of a single URI. So the URL would denote not a unique identifier of the article, but a history state. (Coming to that article from that category)

The `undefined` thing does pose an issue though. Maybe categoryId could be changed to `parentId` and in case of search results, it'd have a value of `search`, otherwise it would have the `categoryId`.

Another alternative is inspecting `history.entries`. In that case, the URL wouldn't have any info about the category.

kantord:
@git-toni 

> you will have access to all the past entries of history - particularly at props.history.entries of that component.

As @jaroslav-kubicek pointed it, it might be harmful to assume that there will be any history entries at all.  If, in a future version, we have to use a magic URL to take the user to a specific article for example, history will be empty.

src/ContentPage/index.js
git-toni: >  So the URL would denote not a unique identifier of the article, but a history state

Great point! :) However if that URI is going to denote the history, how does the current implementation cover the case when an article has been entered after more than one category screen(eg. at the 2nd categories screen)?

Imho breadcrumbs should just be a reflection of the browser history, but I also like your idea of having state embedded in the URI somehow.

kantord: It's @jaroslav-kubicek's idea :-P At first I also implemented with history.

The nested category case is solved, because categories have only one parent. (It's a tree if you don't count the articles). It can be directly fetched from graphql as already done on the category list screen

git-toni:
@kantord 
> If, in a future version, we have to use a magic URL to take the user to a specific article for example, history will be empty.

That's also a concern indeed. I imagine we need to clarify how the magic links will be created - eg. will it have the user history embedded? My initial idea regarding magic links was that they would be built on backend looking something like `/smart-faq?info=4rWnKLaW_3e0R`, and that we could decode that hash and read data about the route. But that might not be enough? :sweat_smile:

kantord:
@git-toni 

Yes but in that case, I think it'll still be easier to use an URL internally, because:
- we are already using a router based on URL
- If we used history inspection, then we'd have to reconstruct history based on that API

However, if we use URL for the same purpose, just by pushing a new URL to history, the state reconstruction will already be implemented.

kantord:
@jaroslav-kubicek @git-toni looks like I finally solved the issues and now the solution correctly handles search + has all path steps. I also resolved the conflicts

git-toni:
Some relay issues apparently, can you confirm? :broken_heart: 

image

kantord:
@git-toni hmm, seems to work fine here. Does `yarn relay` change anything?

git-toni:
I'm talking about the deployed version to now.sh

kantord:
hmmm weird :-D Any ideas?

src/StaticFAQ/FAQArticleDetail.js
```diff
@@ -44,21 +65,45 @@ class FAQArticleDetail extends React.Component<Props> {
...
-          query={queryFAQArticleDetail}
-          variables={{ id: articleId }}
+          query={
+            categoryId === 'search'
+              ? queryFAQArticleDetailSearchResult
+              : queryFAQArticleDetail
+          }
+          variables={{
+            id: articleId,
+            ...(categoryId === 'search' ? {} : { category_id: categoryId }),
```
git-toni: This is repeating the same conditional as in L99. Can you just check `if(categoriId==='search')` *once* and then throw one `QueryRenderer` or another depending on the case. Just like in [here](https://github.com/kiwicom/smart-faq/blob/master/src/StaticFAQ/FAQCategoryList.js#L157)

git-toni:
Great job!

PR #145, opened 2018-05-04
Automatically redirect logged in user

### LINKED ISSUE #137: Redirects to booking page automatically
**Acceptance criteria**

- When the smart FAQ is opened and the user is logged in on the frontend, automatically redirect to booking page (so intro page with two buttons is not shown)
- For non-logged in user, it works as before

kantord:
<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-jssetcawnv.now.sh
</url>
</url>

fixes #137

### DIFF
--- a/src/IntroPage/index.js
+++ b/src/IntroPage/index.js
@@ -1,7 +1,6 @@
 // @flow
 
 import * as React from 'react';
-import { Redirect } from 'react-router-dom';
 import css from 'styled-jsx/css';
 import { Typography } from '@kiwicom/orbit-components';
 
@@ -11,6 +10,7 @@ import CloseButton from './../common/CloseButton';
 import { withUser } from '../context/User';
 import type { User } from '../types';
 import FullFAQLink from '../common/FullFAQLink';
+import redirectsLoggedIn from '../common/redirectsLoggedIn';
 
 const style = css`
   .Intro {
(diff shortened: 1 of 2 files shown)


git-toni:
Please add the issue that this is fixing, I believe it's #137 in this case?

Also, IMHO, while this will work well I believe it's worth to think about making a HOC out of this logc, something like `requireLoggedOut`. Then export a `requireLoggedOut`-decorated `IntroPage` component, will will redirect upon mounting if the user is logged-in.

This will allow us to apply the same logic to other components/screens that only should be rendered when there is no session present.

git-toni:
@kantord Agree `requireLoggedOut` is awkward :sweat_smile: The name was just a suggestion, also `loggedOutPage` or sth alike would convey the point. 

Not sure I understand your login page suggestion, what would it do exactly? Check that it's logged-in or logged-out? Redirect where? Or parameterized redirection?

kantord:
Well, normally you'd see something like `requireLoggedIn` which would redirect to a login page if the user is not logged in, then the login page would redirect back to the original page after login.

In our case we have something similar, but it's more a switch, because we don't *require* the user to log in.

Eh. That also sounds awkward and confusing... Actually I'm content with the current solution :-D

git-toni:
Yes but the requirement for this page is for the user to be *logged out*, not logged-in.

I think this same logic(ie. requiring logged-out to be rendered) applies to other screens as well, so making an HOC out of it feels better.

kantord:
hmm, login is required for the upcoming booking part only, though, isn't it?

git-toni:
Please ignore what I just said, ofc we are talking about `requireLoggedIn` not  `loggedOut`. 

I was thinking about a completely different topic :smile_cat:

kantord:
Ok now?

PR #146, opened 2018-05-04
Refactor booking to support all types

### LINKED ISSUE #112: Upcoming booking shows incorrect data
How it should work correctly:

- Outbound is all legs that have `isReturn` as false
- Inbound is all legs that are related to return flight - `isReturn` is true
- On top of the `ONE_WAY` and `RETURN` flight bookings, there is a third option - `MULTICITY` (doesn't have designs yet)
- Show only outbound if booking is `ONE_WAY` - there are one or more Legs and none of them has `isReturn` set to true
- Inbound consists of all legs with `isReturn` set to true - these are return fligts
- If there is more than one Leg for inbound/outbound, show logo as the combination of up to 4 airlines logos (just like on the frontend, there is new orbit component for that)
- Upcoming booking is booking with the `Leg` that has departure time nearest to the current time & in future (so it can't be decided upon only one Leg, but it needs to be iterated over the all Legs in the list to find the correct booking) - might be worth to create GraphQL query for that, something like `upcomingBooking()`

jaroslav-kubicek:
Closes #112 

TODO

- [x] airlines logo needs to be fixed
- [x] render header properly
~- [ ] render past booking if future not available~<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-dzlicfzjnd.now.sh
</url>
</url>


src/routeDefinitions.js
```diff
@@ -9,7 +9,7 @@ export default {
   FAQ_ARTICLE: '/static-faq/article',
   KIWI_LOGIN: '/kiwi-login',
   FORGOTTEN_PASSWORD: '/forgotten-password',
-  UPCOMING_BOOKING: '/upcoming-booking',
+  NEAREST_BOOKING: '/upcoming-booking',
```
git-toni: I think you should rename the actual path to `/nearest-booking` as well.

src/SingleBookingPage/NearestBooking.js
```diff
@@ -0,0 +1,202 @@
...
+        if (!bookingEdge || !bookingEdge.node) {
+          return null;
+        }
+
+        const booking = bookingEdge.node;
+        const delta = this.getDate(booking) - new Date();
+
+        return { delta, booking: booking };
+      })
+      .filter(Boolean);
```
git-toni: Can't we `filter` here for both `delta > 0` AND non-null? This way you can save the filter at Line93.

src/SingleBookingPage/NearestBooking.js
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        if (nearestBooking.delta > bookingWithDelta.delta) {
+          return bookingWithDelta;
+        }
+
+        return nearestBooking;
+      }, null);
+
+    if (nearestUpcomingBooking) {
```
git-toni: We can probably just cover from L106 to L111 with `return nearestUpcomingBooking` or if `null` is specifically desired `return nearestUpcomingBooking || null`

src/SingleBookingPage/NearestBooking.js
```diff
@@ -0,0 +1,202 @@
...
+      return nearestUpcomingBooking;
+    }
+
+    // T.O.D.O. return nearest past booking (let's wait for another GraphQL update)
```
git-toni: Do we really need to return the nearest past booking? Product question.

src/SingleBookingPage/bookingItem/AccordionHeader.js
```diff
@@ -0,0 +1,24 @@
+// @flow
```
git-toni: What about renaming this component and file to `AccordionTitle` or sth similar? IMHO the "header" and "body" of the accordion should be the visible part when untoggled and toggled, respectively.

src/SingleBookingPage/NearestBooking.js
jaroslav-kubicek: Yes

src/SingleBookingPage/bookingItem/AccordionHeader.js
jaroslav-kubicek: Ok, makes sense

src/SingleBookingPage/bookingItem/Header.js
```diff
@@ -0,0 +1,72 @@
...
+const Header = ({ booking }: Props) => {
+  const status = booking.status && bookingStatus[booking.status];
+
+  return (
+    <div>
+      <div className="headerAbove">
+        <div>
+          {booking.databaseId && (
+            <Typography type="secondary">
+              Upcoming trip # {booking.databaseId}
```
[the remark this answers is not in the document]
jaroslav-kubicek: Yes, I forgot. Original function can’t be used since it reversed the number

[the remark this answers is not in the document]
git-toni: Doesn't it reverse it back? 3-unit chunks should start from the end of the number string

src/SingleBookingPage/NearestBooking.js
jaroslav-kubicek: flow unfortunately understand only `.filter(Boolean)`

src/SingleBookingPage/NearestBooking.js
jaroslav-kubicek: flow unfortunately understand only .filter(Boolean)

src/routeDefinitions.js
jaroslav-kubicek: fixed

src/SingleBookingPage/bookingTypes/OneWay.js
```diff
@@ -0,0 +1,30 @@
...
+import Accordion from '../bookingItem/Accordion';
+import type { OneWay_booking } from './__generated__/OneWay_booking.graphql';
+
+type Props = {|
+  booking: OneWay_booking,
+|};
+
+const OneWay = ({ booking }: Props) => (
+  <div>
+    <AccordionHeader title="Trip" />
```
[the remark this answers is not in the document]
jaroslav-kubicek: aye, changed to `Outbound`

src/SingleBookingPage/bookingItem/Header.js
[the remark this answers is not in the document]
jaroslav-kubicek: no, it was really reversed,

IMHO it's good first issue: https://github.com/kiwicom/smart-faq/issues/150

jaroslav-kubicek:
All should be fixed or resolved, cc @git-toni

git-toni:
Great job, thx!

PR #153, opened 2018-05-09
Update orbit components

jalbertsr:
New icons mainly


src/StaticFAQ/ArticleFeedback/ScreenInput.js
```diff
@@ -88,8 +88,14 @@ class ScreenInput extends React.Component<Props, State> {
         backgroundColor="#f5f7f9"
       >
         <form onSubmit={this.handleSubmit}>
-          <div className="close-icon">
-            <Close fill="#bac7d5" height="12" onClick={this.closeScreen} />
+          <div
+            className="close-icon"
+            onClick={this.closeScreen}
+            onKeyUp={this.closeScreen}
```
git-toni: Maybe we can just `onKeyUp={null}`? This would clarify that this listener is added just in order to pass lint rules and serves no actual purpose.

src/StaticFAQ/ArticleFeedback/ScreenThankyou.js
```diff
@@ -37,12 +37,14 @@ const ScreenThankyou = (props: Props) => {
...
-        <div className="close-icon">
-          <Close
-            fill="#bac7d5"
-            height="12"
-            onClick={() => props.changeScreen(screensList.INITIAL)}
-          />
+        <div
+          className="close-icon"
+          onClick={() => props.changeScreen(screensList.INITIAL)}
+          onKeyUp={() => props.changeScreen(screensList.INITIAL)}
```
git-toni: Maybe we can just `onKeyUp={null}`? This would clarify that this listener is added just in order to pass lint rules and serves no actual purpose.

Issue #140, opened 2018-05-04
formatDate util function should work with current locales


jaroslav-kubicek:
`new Date(dateString).toLocaleDateString('en-US', options);` => locale should not be harcoded to 'en-US'. Instead, it should work with current locale

PR #154, opened 2018-05-10
formatDate util function should work with current locales

edgarordonez:
Closes [#140](https://github.com/kiwicom/smart-faq/issues/140)

Changes:
· Added LanguageContext to get the current locale
· Refactored the helper to accept a new parameter and return the date with the correct format by locale.<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-oupfpuzvmh.now.sh
</url>
</url>

### DIFF
--- a/src/SingleBookingPage/bookingItem/Accordion.js
+++ b/src/SingleBookingPage/bookingItem/Accordion.js
@@ -7,7 +7,7 @@ import idx from 'idx';
 import { Typography } from '@kiwicom/orbit-components';
 import { FlightDirect, ChevronDown } from '@kiwicom/orbit-components/lib/icons';
 
-import { formatDate } from '../../helpers/dateUtils';
+import { FormatDate } from '../../helpers/dateUtils';
 import { Box } from '../../common';
 import CarrierLogoWrapper from './CarrierLogoWrapper';
 import type { AccordionTripSummary_trip } from './__generated__/AccordionTripSummary_trip.graphql';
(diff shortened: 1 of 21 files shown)


src/helpers/dateUtils.js
```diff
@@ -2,9 +2,14 @@
 
 export const URGENCY_THRESHOLD = 48;
 
-export const formatDate = (dateString: string) => {
+export const formatDate = (
+  dateString: string,
+  localeUnderscore: string = 'en-EN',
```
git-toni: `en-US`

git-toni:
`Close` ---> `Closes`  :slightly_smiling_face:

src/SingleBookingPage/bookingItem/Accordion.js
```diff
@@ -61,43 +63,51 @@ const Accordion = ({ trip }: Props) => {
...
-                  <span className="inline-icon">
-                    <AirplaneRight />
-                  </span>
-                </span>
-                <span className="location">
-                  <Typography variant="bold" type="primary">
-                    {arrivalCity}&nbsp;({arrivalCode})
-                  </Typography>
-                </span>
+    <LanguageContext.Consumer>
```
git-toni: Maybe making the whole `Accordion` a consumer of Language is overkill. Also other components will most certainly need to format dates.

We could change the approach here and make our utils function, ie. `formatDate` the `Consumer` of our context, since this is where it's needed after all. What do you think? :slightly_smiling_face:

edgarordonez: Yes, I have thought about it, but I wasn't sure about the better approach if coupling the helper with `LanguageContext` or import this in all components when we need. But if we have others components, I agree with you. I'm going to change it. 👨‍💻

src/helpers/dateUtils.js
edgarordonez: Yes, It's a mistake. I'll change it also. Thanks!

PR #155, opened 2018-05-10
Implement responsive design for mobile

### LINKED ISSUE #135: Implement responsive designs for static FAQs
**Acceptance criteria**

- Show smart FAQ on frontend when accessing Kiwi page from mobile device
- Static FAQs look according to designs

https://zpl.io/2jldrrr

kantord:
fixes #135
<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-wiqsqfolye.now.sh
</url>
</url>
</url>
</url>
</url>
</url>
</url>

### DIFF
--- a/src/ContentHeader/index.js
+++ b/src/ContentHeader/index.js
@@ -44,6 +44,7 @@ const loggedInStyle = css`
     font-size: 28px;
     font-weight: bold;
     color: #171b1e;
+    pointer-events: none;
   }
   div.loggedIn {
     display: flex;
@@ -77,6 +78,7 @@ const loggedOutStyle = css`
     font-size: 28px;
     font-weight: bold;
     color: #171b1e;
+    pointer-events: none;
   }
   div.loggedOut {
     display: flex;
@@ -151,9 +153,9 @@ const renderLoggedOut = (
     <div className="loggedOut">
       <div className="signInOrBack">
         {hasCategory || isArticle ? (
-          <span className="backButton">
+          <div className="backButton">
             <BackButton text={comesFromSearch ? 'Search' : 'Back'} />
-          </span>
+          </div>
         ) : (
           <Link
             to={routeDefinitions.SIGN_IN}
--- a/src/ContentHeader/index.js
+++ b/src/ContentHeader/index.js
@@ -14,7 +14,7 @@ import type { User } from '../types';
 import FullFAQLink from '../common/FullFAQLink';
 
 const style = css`
-  div.logged-out {
+  div.loggedOut {
     display: flex;
     justify-content: space-between;
     padding: 15px 122px 15px 40px;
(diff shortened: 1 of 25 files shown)


src/ContentHeader/index.js
```diff
@@ -14,6 +14,11 @@ import type { User } from '../types';
 import FullFAQLink from '../common/FullFAQLink';
 
 const style = css`
+  div.logged-out {
```
jaroslav-kubicek: please use `camelCase` for classes

src/ContentHeader/index.js
```diff
@@ -55,6 +60,18 @@ const loggedInStyle = css`
   a.open-icon {
     margin-left: 12px;
   }
+  @media only screen and (min-device-width: 320px) and (max-device-width: 480px) {
+    .desktop-only {
+      display: none !important;
```
jaroslav-kubicek: Might be feasible to define `desktopOnly` and `mobileOnly` globally

jalbertsr:
just two observations:

help not centred: 
screenshot

text area too big:
screenshot

kantord:
@jalbertsr they should be resolved now :-)

kantord:
@git-toni It took a while until I figured out the issue with the 'Back' button... Turns out an invisible part of the help header was overlapping it :man_facepalming:

With that, it seems everything is resolved :-)

jalbertsr:
when using it on desktop now I see this: 

gif

kantord:
Oh!

kantord:
At least apparently this only happens when logged out. ( :roll_eyes: :joy_cat: )

src/NoBookingPage/index.js
```diff
@@ -15,6 +15,7 @@ import routeDefinitions from '../routeDefinitions';
 
 const style = css`
   div.NoBookingPage {
+    max-width: 480px;
```
kantord: Apparently this fixed the issue (I checked with mobile and desktop, logged in and logged out).

But if I put

```
width: 480px;
```

it breaks other stuff, for example the breadcrumbs :joy:  Somebody explain why that is....

kantord:
Looks like now.sh passed too :-)

kantord:
And fyi I removed the Sign In buttons. Martin confirmed they have to be removed https://skypicker.slack.com/archives/C9K87QN6Q/p1526643257000054

src/ContentPage/index.js
```diff
@@ -25,6 +25,21 @@ const style = css`
   div.BookingInfo {
     width: 548px;
   }
+
+  @media only screen and (min-device-width: 320px) and (max-device-width: 480px) {
```
jaroslav-kubicek: what about using `min-width` and `max-width`?

kantord: You mean `min-width` instead of `min-device-width` in every media query (same for `max-.*`)?

kantord: I used `min-device-width` because that's what's used all over the project, however I think this is a very good idea and I can't think of any reason why we should use `-device-`, so I replaced it project-wide.

I think this pull request should be squashed, but maybe I wouldn't squash the project-wide replacement into the same one commit. What do you think? Maybe I should just move that commit to another pull request.

jaroslav-kubicek: IMHO it's fine to squash

jaroslav-kubicek:
Ok, it looks really cool, good job man! 👍

PR #159, opened 2018-05-11
added a function to split the ID each 3 digits and return formatted booking id

### LINKED ISSUE #150: Format booking number according to desings
Acceptance criteria
- booking number is divided by space into groups of three digits, just like CZ format for thousands 

screen shot 2018-05-09 at 13 41 25

Related to #146

edgarordonez:
closes #150 

· Added a `formatBookingId` function to receive the ID and return this formatted.<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-qhrtkrvikd.now.sh
</url>
</url>

### DIFF
--- a/src/SingleBookingPage/bookingItem/Header.js
+++ b/src/SingleBookingPage/bookingItem/Header.js
@@ -5,16 +5,12 @@ import { graphql, createFragmentContainer } from 'react-relay';
 import { Typography } from '@kiwicom/orbit-components';
 
 import bookingStatus from '../../common/booking/bookingStatuses';
+import formatBookingId from '../../helpers/formatBookingId';
 import type { Header_booking } from './__generated__/Header_booking.graphql';
 
 type Props = {|
   booking: Header_booking,
 |};
-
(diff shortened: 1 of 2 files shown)


src/SingleBookingPage/bookingItem/Header.js
```diff
@@ -11,6 +11,11 @@ type Props = {|
   booking: Header_booking,
 |};
 
+const formatBookingId = (databaseId: number): string => {
```
jalbertsr: you can move this function to `/helpers` we'll need it in more places

edgarordonez: Yes, another name would be better, or it's okay so? I also thought to add another parameter to decide the "slices". What do you think?

Thanks @jalbertsr

jalbertsr:
@MRTNKCR @mjancik the format should be `# 6 704 418` or is the screenshot correct?

Screenshot:
screenshot

mjancik:
It should be like this: ` # 6 704 418`

PR #160, opened 2018-05-11
99 add flight accordion

### LINKED ISSUE #99: Add Flight info dropdown/accordion
The Upcoming booking page is ready and already has an (incomplete) Accordion component. So we can add the flight info once the design is added to Zeplin. @MRTNKCR @mjancik

Marvel link: https://marvelapp.com/7iihb90/screen/39641486

**Update:** zeplin link: https://zpl.io/adMXoOp

git-toni:
Closes #99 
<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-kfogismpnp.now.sh
</url>
</url>
</url>

### DIFF
--- a/src/SingleBookingPage/bookingItem/AccordionBody.js
+++ b/src/SingleBookingPage/bookingItem/AccordionBody.js
@@ -14,22 +14,14 @@ type Props = {|
 class AccordionBody extends React.Component<Props> {
   renderLegs = () => {
     const { legs } = this.props;
-    const lastLeg = legs.slice(-1)[0];
-    if (legs.length > 1) {
-      return legs
-        .slice(0, -1)
-        .map((l, li) => (
-          <AccordionBodyLeg
-            key={l.flightNumber}
-            leg={l}
-            nextLeg={legs[li + 1]}
-          />
-        ))
-        .concat(
-          <AccordionBodyLastLeg key={lastLeg.flightNumber} leg={lastLeg} />,
-        );
-    }
-    return <AccordionBodyLastLeg key={lastLeg.flightNumber} leg={lastLeg} />;
+    return legs.map((l, li) => {
+      if (li === legs.length - 1) {
+        return <AccordionBodyLastLeg key={l.flightNumber} leg={l} />;
+      }
+      return (
+        <AccordionBodyLeg key={l.flightNumber} leg={l} nextLeg={legs[li + 1]} />
+      );
+    });
   };
   render() {
     return (
--- a/src/SingleBookingPage/bookingItem/AccordionBody.js
+++ b/src/SingleBookingPage/bookingItem/AccordionBody.js
@@ -1,0 +1,67 @@
+// @flow
+
+import * as React from 'react';
+import { createFragmentContainer, graphql } from 'react-relay';
+
+import AccordionBodyLeg from './AccordionBodyLeg';
+import AccordionBodyLastLeg from './AccordionBodyLastLeg';
+import type { AccordionBody_legs } from './__generated__/AccordionBody_legs.graphql';
+
+type Props = {|
+  legs: AccordionBody_legs,
+|};
+
+class AccordionBody extends React.Component<Props> {
+  renderLegs = () => {
+    const { legs } = this.props;
+    const lastLeg = legs.slice(-1)[0];
+    if (legs.length > 1) {
+      return legs
+        .slice(0, -1)
+        .map((l, li) => (
+          <AccordionBodyLeg
+            key={l.flightNumber}
+            leg={l}
+            nextLeg={legs[li + 1]}
+          />
+        ))
+        .concat(
+          <AccordionBodyLastLeg key={lastLeg.flightNumber} leg={lastLeg} />,
+        );
+    }
+    return <AccordionBodyLastLeg key={lastLeg.flightNumber} leg={lastLeg} />;
+  };
+  render() {
+    return (
+      <React.Fragment>
+        <hr />
+        <div className="allLegs">{this.renderLegs()}</div>
+        <style jsx>
+          {`
+            div.allLegs {
+              padding: 16px 53px;
+              padding-bottom: 40px;
+            }
+            hr {
+              margin-top: 16px;
+              border: none;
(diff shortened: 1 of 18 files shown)


src/SingleBookingPage/bookingItem/AccordionHeader.js
```diff
@@ -0,0 +1,98 @@
...
+    justify-content: center;
+    cursor: pointer;
+  }
+  .itinerary .location {
+    margin-right: 4px;
+  }
+`;
+type Props = {|
+  trip: AccordionTripSummary_trip,
+  toggle: () => void,
```
jaroslav-kubicek: IMHO the whole AccordionHeader should be clickable and this component should accept expanded true/false for chevron icon

git-toni: I agree, whole AccordionHeader would be more comfortable. However let's make sure with @MRTNKCR @mjancik's opinoin :slightly_smiling_face: 

> this component should accept expanded true/false for chevron icon

Do you mean the parent `Accordion` component should hold the `toggled/untoggled` state and pass it as props to `AccordionHeader`? If so, I agree, it makes sense conceptually for `Acoordion` to hold the state.

jaroslav-kubicek: > I agree, whole AccordionHeader would be more comfortable. However let's make sure with @MRTNKCR @mjancik's opinoin

yes, make the whole header clickable

then you don't need to pass down `toggle` callback and `expanded=true/false` should be enough

src/SingleBookingPage/bookingItem/AccordionBody.js
```diff
@@ -0,0 +1,67 @@
...
+
+type Props = {|
+  legs: AccordionBody_legs,
+|};
+
+class AccordionBody extends React.Component<Props> {
+  renderLegs = () => {
+    const { legs } = this.props;
+    const lastLeg = legs.slice(-1)[0];
+    if (legs.length > 1) {
```
jaroslav-kubicek: what about just using map function `.map(leg, index, legs) => { 
  if (index === legs.length - 1) {
    return <AccordionBodyLastLeg />; 
  }

  return <AccordionBodyLeg />
}`

src/SingleBookingPage/bookingItem/AccordionBodyLastLeg.js
```diff
@@ -0,0 +1,75 @@
...
+  leg: AccordionBodyLastLeg_leg,
+|};
+
+const RawAccordionLastLeg = (props: LastLegProps) => {
+  const { leg } = props;
+  const initialDeparture = idx(leg.departure, _ => _.localTime) || '';
+  return (
+    <React.Fragment>
+      <div className="lastLeg">
+        <div className="calendar-icon">
```
jaroslav-kubicek: please use consistent names for classes `lastLeg` vs `calendar-icon`

src/SingleBookingPage/bookingItem/AccordionBodyLeg.js
```diff
@@ -0,0 +1,124 @@
+// @flow
```
jaroslav-kubicek: I would suggest one more thing - let's merge these two components together, `AccordionBodyLastLeg` and `AccordionBodyLastLeg` and pass nextLeg={null} for the last one

btw `Kiwi.com Guarantee` needs to be displayed based on data from GraphQL (new field needs to be added) - so let's just render it everytime now

src/SingleBookingPage/bookingItem/AccordionHeader.js
git-toni: > then you don't need to pass down toggle callback

Then who is the one toggling the `expanded` state?

src/SingleBookingPage/bookingItem/AccordionBodyLeg.js
git-toni: > then you don't need to pass down toggle callback

I think the data extracted via their fragments is quite different from one case to the other. If we want to be strict about relay, shouldn't we keep them separate as to not over-request/feed data to the component? (definitely not a critical case, just saying :slightly_smiling_face: )

src/helpers/dateUtils.js
```diff
@@ -35,3 +35,14 @@ export const formatCountDown = (hoursLeft: number) => {
...
   return `${ndays} days`;
 };
+
+export const formatTimeDuration = (mins: number) => {
+  const nhours = Math.floor(mins / 60);
+  const nmins = mins - nhours * 60;
+  return nhours ? `${nhours}h ${nmins}min` : `${nmins}min`;
+};
+
+export const formatHour = (date: string, is12Hour: boolean = true) => {
```
jaroslav-kubicek: btw looks like `is12Hour` param is not needed and it's true always

src/helpers/__tests__/dateUtils.test.js
```diff
@@ -29,4 +29,18 @@ describe('date utils', () => {
...
+    it('should format duration', () => {
+      expect(dateUtils.formatTimeDuration(90)).toBe('1h 30min');
+      expect(dateUtils.formatTimeDuration(32)).toBe('32min');
+      expect(dateUtils.formatTimeDuration(191)).toBe('3h 11min');
+    });
+  });
+  describe('formatHour', () => {
+    it('should format duration', () => {
+      const myDate = '2018-05-14T13:30:16.905Z';
+      expect(dateUtils.formatHour(myDate)).toBe('3:30 PM');
```
jaroslav-kubicek: this isn't correct, it should display `1:30 PM`

CI fails correcly
toLocaleString needs to take `timeZone: 'UTC'` as part of the options

git-toni: Yep, you're right. I was making sure all graphql times are UTC-timezoned, and that is the case. Fixing this now, thx.

src/SingleBookingPage/bookingItem/AccordionBodyLeg.js
jaroslav-kubicek: Now I don't know how it's related to relay 🤔

PR #161, opened 2018-05-11
all booking page

### LINKED ISSUE #98: "Choose another booking" page
We can have the All Upcoming bookings screen already. As discussed we need the design at Zeplin @MRTNKCR @mjancik :slightly_smiling_face:

jalbertsr:
Close #98 

Waiting for https://github.com/kiwicom/graphql/pull/145 to be merge so I can use carrier logos easier and better

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1,5 +1,5 @@
 # source: https://graphql.kiwi.com/
-# timestamp: Mon May 14 2018 11:24:48 GMT+0200 (CEST)
+# timestamp: Tue May 15 2018 11:49:29 GMT+0200 (CEST)
 
 schema {
   query: RootQuery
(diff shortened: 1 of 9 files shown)


src/AllBookings/index.js
```diff
@@ -0,0 +1,129 @@
...
+      return <BookingError />;
+    }
+
+    const edges = idx(props, _ => _.allBookings.edges) || [];
+    const bookings = edges.map(edge => edge.node);
+
+    return bookings.length ? this.renderCardsByType(bookings) : <Loader />;
+  };
+
+  renderAllBookings = (only: string) => (
```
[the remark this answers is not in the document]
mrtnzlml: Inspiration: https://github.com/kiwicom/mobile/blob/5074b23d8c0b59c6a9b7d3514b5eac7b9a4c09f1/app/MMB/src/list/Flights.js#L21 🙂

src/AllBookings/BookingTypes/MulticityBooking.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const trip = idx(booking, _ => _.trips[0]) || [];
+  const departureDate = idx(booking, _ => _.start.time) || '';
+  const origin = idx(booking, _ => _.start.airport.city.name) || '';
+  const IATAOrigin = idx(booking, _ => _.start.airport.locationId) || '';
+  const destination = idx(booking, _ => _.end.airport.city.name) || '';
+  const IATADestination = idx(booking, _ => _.end.airport.locationId) || '';
+
+  return (
+    <BookingCard
```
mrtnzlml: `BookingCard` component is pretty long with quite a lot of properties, isn't it? I would suggest to leverage Relay fragments and decompose it a little bit. It's absolutely fine to compose small components with tiny fragments. In fact we have one rule - write fragments only for what your components need. You are fetching here data for `BookingCard` which we consider to be a wrong approach... 🙂

mrtnzlml: BTW: this is how we do it: https://github.com/kiwicom/mobile/blob/5074b23d8c0b59c6a9b7d3514b5eac7b9a4c09f1/app/MMB/src/list/MulticityFlight.js#L28

As you can see we are not even fetching real data there - only spreading the fragments in order to deal with the multiple booking types we have. This greatly simplifies our code because we don't have to repeat ourselves (see how the fragments are sooo similar here)...

PR #165, opened 2018-05-11
Refactor SingleBookingPage to properly show notification & use "nearestBooking" query

### LINKED ISSUE #148: Render past booking if future not available
Related to #146 

Needs new GraphQL query to do it properly

jaroslav-kubicek:
Closes #148 

Changes

- use new "nearestBooking" query to fetch the most relevant booking corrently, without hassle on frontend
- show notification only when the whole trip haven't started yet
<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-jojtdddlyq.now.sh
</url>
</url>


jalbertsr:
I see PR https://github.com/kiwicom/graphql/pull/139 merged, but can't find the `nearestBooking` query on https://kiwi-graphiql.now.sh/ I guess this is why I get this error when trying this PR locally:
screenshot
how frequently graphql proxy is deployed ? Could it have failed?

jalbertsr:
I know it's an orbit thing, but I did [this change](https://github.com/kiwicom/smart-faq/pull/165/files#diff-655541ebd8ebb4ca180124a4e32c119fL50) cause if there is the case that you have same carriers in the array, the `key` react warning will apear, should we change this in orbit ?

screenshot
screenshot

package.json
```diff
@@ -95,6 +95,7 @@
     "i18next": "10.5.0",
     "idx": "2.2.0",
     "isomorphic-fetch": "2.2.1",
+    "luxon": "1.2.0",
```
git-toni: Herewith the end of custom Date/Time helpers :smile:

src/SingleBookingPage/NearestBooking.js
```diff
@@ -13,194 +13,135 @@ import Notification from './bookingItem/Notification';
...
-                background-color: #e8edf1;
-                border: none;
-                color: #46515e;
-                cursor: pointer;
-              }
-            `}
-          </style>
+    const { booking } = this.props;
+    const departure = this.getDepartureByType(booking);
+    const timeDelta = departure
```
git-toni: Could we just `if(!departure) return null` and avoid all the comparisons with `null` and this ternary here?

git-toni: Yep, you're right. 

I'm just wondering, if there's no departure, it means there's no booking to show at all - neither future not past-, isn't it? Maybe we should have a screen for that "edge" case?

jaroslav-kubicek: well, it's theoretical, but GraphQL can return everything but departure date, as all fields are nullable.

jaroslav-kubicek:
> I know it's an orbit thing, but I did this change cause if there is the case that you have same carriers in the array, the key react warning will apear, should we change this in orbit

Yes, this sucks, I would like to fix this by `carriers` field on Booking, I will prepare some update for Single booking page

PR #175, opened 2018-05-15
157 logout stale session

### LINKED ISSUE #157: Refactor our custom QueryRenderer
Repro steps:
- User is logged in, but his session is stale
- GraphQL query return 403 Forbidden

Expected
- 403 is recognized, `onLogout` callback is called, so stale user session is safely destroyed

Actual: just blank error is shown: see the screen

screen shot 2018-05-10 at 16 17 17

git-toni:
Closes #157 

Rather simple approach based on throwing error upon non-200 response instead of refactoring our custom `QueryRenderer`.

Other screens might have to replicate this feature so another option is to move the `onLogout` + redirect logic to an external helper.

### DIFF
--- a/src/ContentPage/index.js
+++ b/src/ContentPage/index.js
@@ -1,8 +1,9 @@
 // @flow
 
-import { Route, Switch, MemoryRouter } from 'react-router-dom';
+import { withRouter, Route, Switch, MemoryRouter } from 'react-router-dom';
 import * as React from 'react';
 import css from 'styled-jsx/css';
+import type { RouterHistory } from 'react-router-dom';
 
 import ContentHeader from '../ContentHeader';
 import StaticFAQ from '../StaticFAQ';
@@ -44,7 +45,7 @@ const style = css`
   }
 `;
 
-const BookingRoutes = (
+const BookingRoutes = (returnToLogin: () => void) => (
   <MemoryRouter
     initialEntries={[routeDefinitions.NEAREST_BOOKING]}
     initialIndex={0}
@@ -53,17 +54,23 @@ const BookingRoutes = (
       <Route
         exact
         path={routeDefinitions.NEAREST_BOOKING}
-        component={NearestBooking}
+        component={props => (
+          <NearestBooking {...props} returnToLogin={returnToLogin} />
+        )}
       />
       <Route
         exact
         path={`${routeDefinitions.SELECTED_BOOKING}/:bookingId`}
-        component={SelectedBooking}
+        component={props => (
+          <SelectedBooking {...props} returnToLogin={returnToLogin} />
+        )}
       />
       <Route
         exact
         path={routeDefinitions.ALL_BOOKINGS}
-        component={AllBookings}
+        component={props => (
+          <AllBookings {...props} returnToLogin={returnToLogin} />
+        )}
       />
     </Switch>
   </MemoryRouter>
@@ -84,12 +91,21 @@ const FAQRoutes = (
     </Switch>
   </MemoryRouter>
 );
-const ContentPage = () => {
+
+type Props = {
+  history: RouterHistory,
+};
+const ContentPage = (props: Props) => {
+  const returnToLogin = () =>
+    props.history.push({
+      pathname: routeDefinitions.SIGN_IN,
+      state: { sessionExpired: true },
+    });
   return (
     <div className="ContentPage">
       <ContentHeader />
       <div className="Body">
-        <div className="BookingInfo">{BookingRoutes}</div>
+        <div className="BookingInfo">{BookingRoutes(returnToLogin)}</div>
         <div className="FAQ">{FAQRoutes}</div>
       </div>
       <style jsx global>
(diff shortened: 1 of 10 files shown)


src/SingleBookingPage/index.js
```diff
@@ -30,6 +36,10 @@ const query = graphql`
 class SingleBookingPage extends React.Component<Props> {
   renderBooking = (renderState: RenderState) => {
     if (renderState.error) {
+      if (renderState.error.message === ERROR_FORBIDDEN) {
+        this.props.onLogout();
```
jaroslav-kubicek: doesn't `onLogout` return promise? it should be awaited in such case

src/relay/netorkErrors.js
```diff
@@ -0,0 +1,3 @@
+// @flow
```
jaroslav-kubicek: typo in filename

jaroslav-kubicek: there is no necessity to have it defined in separate file, it's overkill ==> relay/environment.js

src/relay/environment.js
```diff
@@ -41,6 +42,12 @@ const buildFetchQuery = (
       }),
     });
     const json = await response.json();
+    if (json.errors && json.errors.length > 0) {
+      const errorMsg = json.errors.some(e => e._proxy.statusCode === 403)
+        ? ERROR_FORBIDDEN
+        : '';
+      throw new Error(errorMsg);
```
jaroslav-kubicek: - We should allow partial errors
- throwing error with empty message doesn't make much sense

src/SingleBookingPage/index.js
jaroslav-kubicek: also, probably could be moved to custom QueryRenderer

git-toni:
cc @jaroslav-kubicek

src/context/Language.js
```diff
@@ -5,3 +5,16 @@ import * as React from 'react';
...
 
 export const LanguageContext = React.createContext(defaultLanguage);
+
+export const withDefaultLanguage = <Props>(
+  Component: React.ComponentType<{ defaultLanguage: string } & Props>,
+) =>
+  function withLanguageHOC(props: Props) {
+    return (
+      <LanguageContext.Consumer>
+        {(defaultLanguage: string) => (
```
jaroslav-kubicek: `defaultLanguage` is rather unfortunate name for language prop

src/relay/QueryRenderer.js
```diff
@@ -18,25 +20,41 @@ type Props<RenderProps> = {|
...
+};
+type RenderState<RenderProps> = {
+  props: ?RenderProps,
+  error: ?Error,
+};
+class QueryRenderer extends React.Component<ContextTypes & Props<RenderProps>> {
+  renderContainer = (renderState: RenderState<RenderProps>) => {
+    if (renderState.error) {
+      if (renderState.error.message === ERROR_FORBIDDEN) {
+        this.props.onLogout().then(() => {
```
jaroslav-kubicek: should use async/await instead

src/ContentPage/index.js
```diff
@@ -84,12 +91,21 @@ const FAQRoutes = (
     </Switch>
   </MemoryRouter>
 );
-const ContentPage = () => {
+
+type Props = {
+  history: RouterHistory,
+};
+const ContentPage = (props: Props) => {
+  const returnToLogin = () =>
```
jaroslav-kubicek: this should be moved down to `QueryRenderer` as well, no need to pass it down as prop

src/context/Language.js
git-toni: :smile: I don't find it that crazy offtopic, but you've got a point. What about `language`? :book:

src/relay/QueryRenderer.js
git-toni: How do we do that before they land [async rendering](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) in React?

src/ContentPage/index.js
git-toni: The problem I faced here and why I ended up with this rather ugly solution is that QueryRenderer is instantiated by many different components, and some of the components are not rendered under the "main" MemoryRouter. But I am only interested in the returnToLogin that pushes history to the "main" router(the parent of ContentPage, otherwise I would only redirect to login half of the page, ie. a sub-router like the one of BookingRoutes in Line 48 here. Hence I cannot decorate QueryRenderer with withRouter cause that would not be linked to the desired main MemoryRouter instance.

All of this will be thankfully obsolete with your new proposal for Routing. So I am totally fine having you merge the new routing system first, and then rebasing this PR, however painful. slightly_smiling_face

src/relay/QueryRenderer.js
jaroslav-kubicek: good point 🤔 

maybe 
```
(async () => { 
  await something(); 
})().catch(console.error)
```

src/ContentPage/index.js
jaroslav-kubicek: ok, le's merge it and let's suffer in my PR

PR #198, opened 2018-05-17
fix accordion

### LINKED ISSUE #188: Long Accordion overflows parent container
According to [design](https://app.zeplin.io/project/5a8e8992554d385c76ad5aeb/screen/5aeb0daac11774fb37dd05eb) there shouldn't be scroll but the screen should be "cut" instead. 

However in some cases (like the one below) the flight has so many legs even that single flight doesn't fit in the screen. Should we then add scrolls somewhere? @mjancik @MRTNKCR 

image

### LINKED ISSUE #195: Fix Accordion
Accordion on frontend can't be expanded because there are already global styles for `.Accordion` class setting the fixed height:

```
height: 72px;
max-height: 72px;
```

Solution - use a different class?

jalbertsr:
Close #188 
Close #195 

Manual Deploy LiveURL: https://kiwicomsmart-faq-hfnpahvusi.now.sh<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-eagcbeqsnx.now.sh
</url>
</url>
</url>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # smart-faq
 Smart FAQ
 
-App demo: [https://kiwicom.github.io/smart-faq/](https://kiwicom.github.io/smart-faq/)
+App demo: [https://kiwicom.github.io/smart-faq/](https://kiwicom.github.io/smart-faq/)  
 Frontend staging: [https://introduce-smart-faq.fe.staging.kiwi.com/](https://introduce-smart-faq.fe.staging.kiwi.com/)
 
 ## Translations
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 # smart-faq
 Smart FAQ
 
-App demo: [https://kiwicom.github.io/smart-faq/](https://kiwicom.github.io/smart-faq/)
+App demo: [https://kiwicom.github.io/smart-faq/](https://kiwicom.github.io/smart-faq/)
```
jalbertsr: sorry I couldn't resist 🙂

PR #200, opened 2018-05-18
Collapse multicity trips

### LINKED ISSUE #156: Collapse multicity trips
If multicity booking has more than 2 trips, show only first two with overlay effect and enable to show them all - by clicking on "Show all trips" **(Not Show all flights!)**

screen shot 2018-05-10 at 15 46 38

Zeplin: https://zpl.io/bLw5z8Q

edgarordonez:
Close #156 

- Added a new component to wrap Multicity with overlay effect.
- Changed `NearestBooking` to use `MulticityOverlay`.<br/><br/><br/><url>LiveURL: https://smartfaq-156-collapse-multicity-trips.surge.sh</url>
</url>
</url>

### DIFF
--- a/src/SingleBookingPage/bookingTypes/MulticityOverlay.js
+++ b/src/SingleBookingPage/bookingTypes/MulticityOverlay.js
@@ -1,0 +1,125 @@
+// @flow
+
+import * as React from 'react';
+import { createFragmentContainer, graphql } from 'react-relay';
+
+import Multicity from './Multicity';
+import type { MulticityOverlay_booking } from './__generated__/MulticityOverlay_booking.graphql';
+
+type Props = {|
+  booking: MulticityOverlay_booking,
+|};
+
+type State = {|
+  overlapping: boolean,
+|};
+
+class MulticityOverlay extends React.Component<Props, State> {
+  state = {
+    overlapping: true,
+  };
+
+  toggleOverlapping = () => {
+    this.setState({ overlapping: !this.state.overlapping });
+  };
+
+  render() {
+    const { booking } = this.props;
+    const { trips } = booking;
+
+    return trips && trips.length > 2 ? (
+      <div>
+        <div
+          className={
+            this.state.overlapping
+              ? 'multicityWrapper'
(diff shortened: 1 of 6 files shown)


src/SingleBookingPage/bookingTypes/MulticityOverlay.js
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  render() {
+    const { booking } = this.props;
+
+    return (
+      <div
+        className={
+          this.state.overlapping
+            ? 'multicity-wrapper'
```
jaroslav-kubicek: We should use `.camelCase` for classNames

jaroslav-kubicek:
I found several issues:

screen shot 2018-05-18 at 13 26 18

- Expanded state doesn't work for me - there should be "show less"
- Down show any overlay if there are two trips at most, only if `trips.length > 2`

Relay fragments are set wrong - try running `yarn relay`

Few hints:

if you define fragment for `MulticityOverlay` like this, you can then check the length of the trips in render function:

```
export default createFragmentContainer(
  MulticityOverlay,
  graphql`
    fragment MulticityOverlay_booking on BookingMulticity {
      trips {
        duration
      }
      ...Multicity_booking
    }
  `,
);
```

in SingleBookingPage/index.js file, change the query to (just temporarily for local testing):

```
const query = graphql`
  query SingleBookingPageQuery {
    booking(id: "Qm9va2luZzozNzQ5MDU5") {
      multicity {
        ...NearestBooking_booking
      }
    }
  }
`;
```

and renderBooking function to:

```
  renderBooking = (renderState: RenderState) => {
    if (renderState.error) {
      return <BookingError />;
    }

    if (!renderState.props) {
      return <Loader />;
    }

    const booking = renderState.props.booking;

    if (!booking) {
      return <BookingNotFound />;
    }

    return <NearestBooking booking={booking.multicity} />;
  };
```

you will need to also update Graphql schema, because `duration` field was added recently and it's not in our local snapshot in `schema.graphql`

jaroslav-kubicek:
We should finish this, but there are conflicts

ping @edgarordonez

PR #204, opened 2018-05-18
show appropriate booking header

### LINKED ISSUE #151: Render Booking header according to designs
Based on booking type, render appropriate header for booking:

Return:

screen shot 2018-05-09 at 13 48 21

Multicity:
screen shot 2018-05-09 at 13 48 48

One way - just like return without "and back"

render "Last trip"/"Upcoming trip" as well depending on whether the booking is in future or not.

jalbertsr:
Close #151 <br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-tabyontjld.now.sh
</url>
</url>
</url>
</url>
</url>

### DIFF
--- a/src/SingleBookingPage/bookingItem/BookingHeaders/Multicity.js
+++ b/src/SingleBookingPage/bookingItem/BookingHeaders/Multicity.js
@@ -11,12 +11,13 @@ type Props = {|
 
 const MulticityBookingHeader = ({ bookingHeader }: Props) => {
   const trips = idx(bookingHeader, _ => _.trips) || [];
-  const destiantion = idx(bookingHeader.end, _ => _.airport.city.name) || '';
+  const destination = idx(bookingHeader.end, _ => _.airport.city.name) || '';
   return (
     trips.reduce((acc, trip) => {
       const city = idx(trip, _ => _.departure.airport.city.name) || '';
+      //F.I.X.M.E: new Orbit release should publish new leftArrow icon
       return (acc += `${city} → `);
-    }, '') + destiantion
+    }, '') + destination
   );
 };
 
--- a/src/SingleBookingPage/bookingItem/BookingHeaders/Multicity.js
+++ b/src/SingleBookingPage/bookingItem/BookingHeaders/Multicity.js
@@ -1,0 +1,45 @@
+// @flow
+
+import idx from 'idx';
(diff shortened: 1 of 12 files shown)


src/SingleBookingPage/bookingItem/BookingHeaders/Multicity.js
```diff
@@ -0,0 +1,45 @@
...
+  bookingHeader: Multicity_bookingHeader,
+|};
+
+const MulticityBookingHeader = ({ bookingHeader }: Props) => {
+  const trips = idx(bookingHeader, _ => _.trips) || [];
+  const destiantion = idx(bookingHeader.end, _ => _.airport.city.name) || '';
+  return (
+    trips.reduce((acc, trip) => {
+      const city = idx(trip, _ => _.departure.airport.city.name) || '';
+      return (acc += `${city} → `);
```
jalbertsr: Orbit components doesn't have available yet the `arrowRight` icon, and couldn't `transform: rotate(90deg)` the `arrowUp`, some ideas? or can we leave it like this until new Orbit update ?

kantord: Why did the transform fail? 

We should not wait for orbit components forever, so I guess if there's no other way, we should leave it like this. I know FIXME comments are not exactly a great pattern but in that case maybe I would leave a FIXME comment there.

kantord:
How do you test the special cases (multi-city, one-way)?

src/SingleBookingPage/bookingItem/BookingHeaders/Multicity.js
jalbertsr: I can rotate the `div` wrapping the icon, but not the icon itself, since is a component from Orbit I can't pass classes to it or use classes in it. Is a good suggestion FIXME comment here 👍, thanks.

kantord: hmm is it possible to add an extra div around the icon and rotate that?
Although even if that works it might not be the most elegant solution long term :-D

jalbertsr: I tried it, without success 😞

jaroslav-kubicek:
See https://github.com/kiwicom/smart-faq/issues/228

PR #205, opened 2018-05-18
Add possibility to select booking from ALLBookings page

jaroslav-kubicek:
<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-hreitykkon.now.sh
</url>


kantord:
I suggest using more granular commits and squashing upon merge, because that makes it easier to understand what happened, than one single and long diff output.

kantord:
2018-05-18-162823_333x94_scrot
the icon here seems too big. In the implementation is 24x24. In the design it's 16x16

Although I think 24x24 is a better size :-)

jalbertsr:
@kantord I will resolve this here https://github.com/kiwicom/smart-faq/pull/206 , since is about all booking styles.
I'm with you, I think 24x24 is better size too  :)

PR #211, opened 2018-05-22
added enzyme tests for Login component

dinodsaurus:
<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-dmfxbhqwkp.now.sh
</url>
</url>
</url>

### DIFF
--- a/src/KiwiLogin/__test__/KiwiLogin.test.js
+++ b/src/KiwiLogin/__test__/KiwiLogin.test.js
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { shallow } from 'enzyme';
 
 import { RawKiwiLogin } from '../index';
+import routeDefinitions from '../../routeDefinitions';
 
 describe('KiwiLogin', () => {
   const props = {
@@ -35,7 +36,9 @@ describe('KiwiLogin', () => {
     const Login = wrapper.instance();
     await Login.handleSignIn(e);
     expect(e.preventDefault.mock.calls).toHaveLength(1);
-    expect(Login.props.history.push.mock.calls[0][0]).toBe('/content');
+    expect(Login.props.history.push.mock.calls[0][0]).toBe(
+      routeDefinitions.CONTENT,
+    );
   });
 
   it('should handle error in sign in', async () => {
--- a/src/KiwiLogin/__test__/KiwiLogin.test.js
+++ b/src/KiwiLogin/__test__/KiwiLogin.test.js
@@ -1,23 +1,74 @@
 // @flow
-
 import * as React from 'react';
+import { shallow } from 'enzyme';
+
+import { PureKiwiLogin } from '../index';
 
 describe('KiwiLogin', () => {
-  /*
-  enzyme broken until new release https://github.com/airbnb/enzyme/pull/1513
+  const props = {
+    onLogin: () =>
+      new Promise(resolve => {
+        process.nextTick(() => resolve());
+      }),
+    history: {
+      push: jest.fn(),
+    },
+  };
+  const e = {
+    preventDefault: jest.fn(),
+  };
 
-  const mockStore = configureStore();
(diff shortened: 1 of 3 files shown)


src/KiwiLogin/index.js
```diff
@@ -123,7 +123,7 @@ type State = {|
   showError: boolean,
 |};
 
-class KiwiLogin extends React.Component<Props, State> {
+export class PureKiwiLogin extends React.Component<Props, State> {
```
dinodsaurus: We can discuss about naming of the pure components which are exported without HOC stuff

dinodsaurus: I saw you call it `RawComponent,` so I will follow that naming convention ;)

src/KiwiLogin/__test__/KiwiLogin.test.js
```diff
@@ -1,23 +1,68 @@
...
+    });
+
+    expect(Login.state).toMatchSnapshot();
+  });
+
+  it('should handle sign in', async () => {
+    const Login = wrapper.instance();
+    await Login.handleSignIn(e);
+    expect(e.preventDefault.mock.calls).toHaveLength(1);
+    expect(Login.props.history.push.mock.calls[0][0]).toBe('/content');
```
git-toni: `/content` ----> `routeDefinitions.CONTENT` from `/src/routeDefinitions`.

src/KiwiLogin/__test__/KiwiLogin.test.js
```diff
@@ -1,23 +1,68 @@
...
-      <MemoryRouter initialEntries={[{ key: 'testKey' }]}>
-        <KiwiLogin store={store} />
-      </MemoryRouter>
-    </CloseContext.Provider>
-  );
-  */
+  const props = {
+    onLogin: () =>
+      new Promise(resolve => {
+        process.nextTick(() => resolve());
```
git-toni: Just curious, why not just `new Promise(resolve => resolve())`, wouldn't it end up basically at the same queue position? :slightly_smiling_face:  Idem for Line 51 below.

src/KiwiLogin/__test__/KiwiLogin.test.js
dinodsaurus: I personally like to have stuff hardcoded in tests, so they fail if I make some changes. But if you think this is a better approach I don't mind changing it 👍

git-toni: But aren't you testing that the history was pushed **exactly** to `routeDefinitions.CONTENT`, as expected behavior from the component?

src/KiwiLogin/__test__/KiwiLogin.test.js
dinodsaurus: There is no specific reason behind this, I just always do it like this..
and now when you asked I realized it doesnt make sense 😂
I think the cleanest approach would be `async () => {}`

git-toni: Haha, fair enough. In your defense, jest's docs also use `nextTick` thing. 

`async () => {}` is nice but isn't `onLogin` specifically a `Promise` as received by props? https://github.com/kiwicom/smart-faq/blob/6e8fe3ad41bf65ab01ab47611e51ac99cb69cdec/src/types.js

src/KiwiLogin/__test__/KiwiLogin.test.js
dinodsaurus: yep, makes sense here 👍

src/KiwiLogin/__test__/KiwiLogin.test.js
git-toni: Ofc, please ignore my question. It should be totally equivalent :sweat_smile:

PR #212, opened 2018-05-22
Intropage tests

dinodsaurus:
<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-hphmebkrrc.now.sh
</url>
</url>

### DIFF
--- a/src/IntroPage/IntroPage.stories.js
+++ b/src/IntroPage/IntroPage.stories.js
@@ -1,0 +1,11 @@
+// @flow
+
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import StoryRouter from 'storybook-router';
+
+import IntroPage from './';
+
+storiesOf('IntroPage', module)
+  .addDecorator(StoryRouter())
+  .add('Default', () => <IntroPage />);
--- a/src/IntroPage/__tests__/IntroPage.test.js
+++ b/src/IntroPage/__tests__/IntroPage.test.js
@@ -1,0 +1,19 @@
+// @flow
+import * as React from 'react';
+import { shallow } from 'enzyme';
+
+import { PureIntro } from '../index';
+import { user } from '../../helpers/defaultTestProps';
+
+describe('IntroPage', () => {
+  const props = {
+    user,
+    history: {
+      push: jest.fn(),
+    },
+  };
+  it('should match snapshot', () => {
+    const wrapper = shallow(<PureIntro {...props} />);
+    expect(wrapper).toMatchSnapshot('<IntroPage />');
+  });
+});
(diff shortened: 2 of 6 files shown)


src/IntroPage/IntroPage.stories.js
```diff
@@ -0,0 +1,11 @@
...
+
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import StoryRouter from 'storybook-router';
+
+import IntroPage from './';
+
+storiesOf('IntroPage', module)
+  .addDecorator(StoryRouter())
+  .add('Default', () => <IntroPage />);
```
jalbertsr: I think you noticed that we almost never use storybook even having it installed (shame on us), we use it so little we haven't realized that storybook wasn't working until your (https://github.com/kiwicom/smart-faq/pull/210), but if we start using it more often I think we should use it to show components not "screens", even so there won't be many components since we relay in Orbit Components as ui library.  what do you think ? @jaroslav-kubicek @git-toni @kantord @edgarordonez

dinodsaurus: Yep makes sense.. sorry for the trouble ;)

git-toni: I totally agree with @jalbertsr . Storybook seems like a great way to showcase our more "complex" components such as Accordion, Breadcrumbs, etc.

src/helpers/defaultTestProps.js
```diff
@@ -0,0 +1,7 @@
+// @flow
```
jaroslav-kubicek: is this file necessary? it's used only once

if we need the same in multiple test files, let's hide it in some `__tests__` folder

src/IntroPage/__tests__/IntroPage.test.js
```diff
@@ -0,0 +1,19 @@
...
+describe('IntroPage', () => {
+  const props = {
+    user,
+    history: {
+      push: jest.fn(),
+    },
+  };
+  it('should match snapshot', () => {
+    const wrapper = shallow(<PureIntro {...props} />);
+    expect(wrapper).toMatchSnapshot('<IntroPage />');
```
jaroslav-kubicek: in case of IntroPage ,it  doesn't matter much, but in general, snapshots tests doesn't make sense for big components as they tend to change too often, snapshot is long & noone is paying attention to it.

small, leaf components => shapshots are fine
bigger components => verifying the existence of particular child component with enzyme `find` or simulation of the events usually works better than huge snapshot

src/IntroPage/IntroPage.stories.js
jaroslav-kubicek: I agree as well

src/helpers/defaultTestProps.js
dinodsaurus: I would say this was a premature optimization 😅

PR #215, opened 2018-05-22
added filter for checking if the carrier is allready added

### LINKED ISSUE #201: Each carrier logo should be shown only once
Each carrier should be represented only by one instance of logo, not like this:

screen shot 2018-05-18 at 11 40 10

dinodsaurus:
closes #201 
<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-rqatpngqwo.now.sh
</url>
</url>
</url>

### DIFF
--- a/src/SingleBookingPage/bookingItem/CarrierLogoWrapper.js
+++ b/src/SingleBookingPage/bookingItem/CarrierLogoWrapper.js
@@ -12,11 +12,19 @@ type Props = {
 };
 
 const CarrierLogoWrapper = (props: Props) => {
+  const addedCarriers = {};
   const carriers = (props.legs || [])
-    .map(leg => {
+    .filter(leg => {
       const code = idx(leg, _ => _.airline.code);
+      if (code && !addedCarriers[code]) {
+        addedCarriers[code] = true;
+        return true;
+      }
+      return false;
+    })
+    .map(leg => {
       const name = idx(leg, _ => _.airline.name);
-
+      const code = idx(leg, _ => _.airline.code);
       if (code && name) {
         return { code, name };
       }
--- a/src/SingleBookingPage/bookingItem/CarrierLogoWrapper.js
+++ b/src/SingleBookingPage/bookingItem/CarrierLogoWrapper.js
@@ -12,28 +12,18 @@ type Props = {
 };
 
 const CarrierLogoWrapper = (props: Props) => {
-  const addedCarriers = {};
-  const carriers = (props.legs || [])
-    .filter(leg => {
-      const code = idx(leg, _ => _.airline.code);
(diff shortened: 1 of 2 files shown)


dinodsaurus:
Sorry @edgarordonez I stole this issue from you 🚂

src/SingleBookingPage/bookingItem/CarrierLogoWrapper.js
```diff
@@ -12,11 +12,19 @@ type Props = {
 };
 
 const CarrierLogoWrapper = (props: Props) => {
+  const addedCarriers = {};
   const carriers = (props.legs || [])
-    .map(leg => {
+    .filter(leg => {
```
git-toni: Not a crazy critical edge case that needs great optimization, but what about something like this? I personally find it more clear:
```javascript
const CarrierLogoWrapper = (props: Props) => {
  const carriers = {};
  (props.legs || []).forEach(leg => {
    const code = idx(leg, _ => _.airline.code);
    const name = idx(leg, _ => _.airline.name);
    if (code && name && !carriers[code]) {
      carriers[code] = { code, name };
    }
  });

  return <CarrierLogo size="large" carriers={Object.values(carriers)} />;
};
```

We could make it all within one loop, but `Object.values` is quite explicit :)

dinodsaurus: Looks better then my filter/map/filter maze!

dinodsaurus: The only problem with this approach is flow.. https://github.com/facebook/flow/issues/2221
So I made it like this `const carriersArray = Object.keys(carriers).map(key => carriers[key]);`
Do you maybe have any suggestions?

git-toni: Oh the flow, that great frie-nemy. 
I didn't try, but a mix of your approach and mine maybe? Avoiding the `Object.values` loop and hence super-mega-optimizing this algorithm to a single loop :roll_eyes: 
```javascript
const CarrierLogoWrapper = (props: Props) => {
  const carriers = {};
  const finalCarriers = [];
  (props.legs || []).forEach(leg => {
    const code = idx(leg, _ => _.airline.code);
    const name = idx(leg, _ => _.airline.name);
    if (code && name && !carriers[code]) {
      carriers[code] = true;
      finalCarriers.push({ code, name });
    }
  });

  return <CarrierLogo size="large" carriers={finalCarriers} />;
};
```

dinodsaurus: We have a battle tested algorithm now 😎

PR #222, opened 2018-05-23
Added booking date to booking card

### LINKED ISSUE #189: Add booking date to booking card
once this pr https://github.com/kiwicom/graphql/pull/150 is merged

- Add booking date to booking card instead of *Not Available*

screenshot

edgarordonez:
Close #189 

- I added the booking date to booking card.<br/><br/><br/><url>LiveURL: https://kiwicomsmart-faq-yuuqywonia.now.sh
</url>
</url>

### DIFF
--- a/src/AllBookings/BookingCard/BookingCard.js
+++ b/src/AllBookings/BookingCard/BookingCard.js
@@ -78,6 +78,7 @@ const BookingCard = (props: Props) => {
   const databaseId = idx(props.booking, _ => _.databaseId);
   const passengerCount = idx(props.booking, _ => _.passengerCount);
   const status = props.booking.status && bookingStatus[props.booking.status];
+  const bookingDate = idx(props.booking, _ => _.bookingDate);
 
   return (
     <div className="card">
@@ -101,6 +102,7 @@ const BookingCard = (props: Props) => {
         status={status}
         departure={props.departure}
         passengerCount={passengerCount}
+        bookingDate={bookingDate}
       />
       <style>{styles}</style>
     </div>
@@ -114,6 +116,7 @@ export default createFragmentContainer(
       status
       databaseId
       passengerCount
+      bookingDate
     }
 
     fragment BookingCard_arrival on RouteStop {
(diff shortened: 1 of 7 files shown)


src/AllBookings/BookingCard/BookingCard.js
```diff
@@ -78,6 +78,7 @@ const BookingCard = (props: Props) => {
   const databaseId = idx(props.booking, _ => _.databaseId);
   const passengerCount = idx(props.booking, _ => _.passengerCount);
   const status = props.booking.status && bookingStatus[props.booking.status];
+  const bookingDate = idx(props.booking, _ => _.bookingDate);
```
git-toni: Shouldn't this field `bookingDate`, together with other (unrelated to this PR) fields such as `passengerCount` and `status` be retrieved from `DateAndPassenger`'s fragment? I mean `BookingCard` component doesn't seem to use them.

jalbertsr: yep, that's true, my bad, it can be an issue a part if you want @edgarordonez ?

edgarordonez: Okay, Do I create a new issue, right? @git-toni

git-toni: That would be ideal I guess :slightly_smiling_face:.

I'm also thinking we should evaluate the `null`ability of those fields, and maybe provide `idx(...) | something|`? Except `bookingDate`, which I believe cannot be `null` nor empty being of `GraphQLDate` type?

edgarordonez: Okay, so I'm gonna to create a new issue to solve it. 😃

jaroslav-kubicek: >  Except bookingDate, which I believe cannot be null

can be null as well, nearly everything can be null in GraphQL 💥

src/AllBookings/BookingCard/__test__/DateAndPassenger.test.js
```diff
@@ -16,8 +16,15 @@ const departure = {
 
 const status = bookingStatus['CONFIRMED'];
 
+const bookingDate = '2017-11-23T20:40:00.000Z';
```
jaroslav-kubicek: I would suggest removing these two variables and pass values directly to the component - or at least remove the empty line on line 18, my OCD meter is broken :D

jalbertsr:
with this https://github.com/kiwicom/smart-faq/pull/231 merged we can close this PR

Issue #208, opened 2018-05-22
Refactor usages of React Router & Route in app to support HTML5 history


jaroslav-kubicek:
- There needs to be only one instance of router managing the url
- navigation thought FAQ is reclected in url
- displaying single booking / all bookings is not handled by router - each custoemr has it's own bookings and having unique url would mean more troubles for us

Issue #184, opened 2018-05-16
Login don't work when going back from FAQ


jaroslav-kubicek:
Repro steps
- Click on "I don't have booking"
- Click on "Sign in" in header
- Sign in via email/password

Expected
- Booking page is displayed as usual

Actual
- Blank page is displayed

<img width="836" alt="screen shot 2018-05-16 at 18 57 50" src="https://user-images.githubusercontent.com/2174629/40131922-9a946f42-593b-11e8-810a-a978c6add022.png">

jalbertsr:
Solved by https://github.com/kiwicom/smart-faq/pull/227

PR #227, opened 2018-05-23
Refactor routing

jaroslav-kubicek:
Closes: #208 

- To enable changing the url, we need to have just one Router in the app, more than one creates collisions<br/><br/><br/><url>LiveURL: https://smartfaq-refactor-routing.surge.sh</url>
</url>
</url>
</url>


jalbertsr:
this PR also fixes #184 👌

jalbertsr:
I guess you removed `routeDefinition` to have more "visible" routes?

src/common/responsiveStyleHelperClasses.js
```diff
@@ -2,10 +2,17 @@
...
+  .mobileOnly {
+    display: none;
+  }
+
   @media only screen and (min-width: 320px) and (max-width: 480px) {
     .desktopOnly {
       display: none !important;
     }
+    .mobileOnly {
+      display: initial !important;
```
kantord: This is good, but we have to be careful with `display: initial`, as it might not have the intended effect. (For example it would convert a `div` with `display: flex` into `display: block`, no? ) Unfortunately I cannot think of a nice solution.

jaroslav-kubicek: fixed

src/common/layout/BookingPage.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import * as React from 'react';
+
+import AllBooking from '../../AllBookings';
+import NearestBooking from '../../SingleBookingPage/NearestBooking';
+import SelectedBooking from '../../SingleBookingPage/SelectedBooking';
+import { type BookingStateType } from '../../context/BookingState';
+
+type Props = BookingStateType;
```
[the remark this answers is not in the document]
jaroslav-kubicek: yep

src/context/BookingState.js
```diff
@@ -0,0 +1,55 @@
...
+
+  onClickSelect = (id: string) => {
+    this.setState({ bookingPage: 'SINGLE_BOOKING', selectedBooking: id });
+  };
+
+  render() {
+    const { bookingPage, selectedBooking } = this.state;
+
+    return (
+      <ClickAllBooking.Provider value={this.onClickAllBooking}>
```
git-toni: Do we really need three contexts here? What about renaming the main context to `BookingScreen` or sth. And Provide from that context both the screen info and the switchers?(see Line 41-42).
Eg.

image

Then we could consume from the required ones like `src/SingleBookingPage/bookingItem/Header.js` de-structuring like this:
image

Or even better, exporting a set of HOC from our `BookingScreen`(now BookingState+Clickers) context to decorate the consumer with the necessary props; just like we do with our User context now: `withLogin`, `withLogout`, etc.

git-toni:
This is a remarkable and creative PR, it'll simplify several things. Thanks!

Few observations:
1. (Premature optimization warning) Designs do not specify animations currently for the SingleBooking/AllBookings pages. However if that happens, how will we support screen transition animations? One option would be to [render](https://github.com/kiwicom/smart-faq/pull/130/files#diff-ccbd0d0511602ed850cdb08bd74814f3R38) a `<Switch>` probably and set the `key` of `CSSTransitions` to our `BookingState.bookingPage` field.
2. Already raised by @jalbertsr. What are the issues encountered with a `routeDefinitions` object? I tried to play with it very little and seemed to have no problems for parameterized paths as well. Matter of taste but I prefer to have all paths used by components and tests centralized. :slightly_smiling_face: 
3. One refactor suggestion. Explained in code.

src/context/BookingState.js
jaroslav-kubicek: fixed

jaroslav-kubicek:
> Already raised by @jalbertsr. What are the issues encountered with a routeDefinitions object? I tried to play with it a bit and seemed to have no problems for parameterized paths as well. Matter of taste but I prefer to have all paths used by components and tests centralized. 

Yep, as I said, I didn't like it much, especially during the refactoring, also because of parameters in url: `SOME_URL/:someParam`. But I'm thinking about changing that in next PR so the benefits are kept, while it's easier to use. e.g. to functions:

```FAQ_ARTICLE: (categoryId, articleId) => `/faq/${categoryId}/article/${articleId}````

jaroslav-kubicek:
all issues should be adress by now, cc @git-toni @jalbertsr

PR #242, opened 2018-05-25
Added Cypress in smart-faq and the test "User can search in FAQs" has been implemented

### LINKED ISSUE #194: Implement scenario "User can seach"
```
Given a user is not logged in on the frontend
And he has opened the smart faq
When he writes e.g. "flight" into search input
Then the list of FAQ article previews is shown (FAQ article preview doesn't have arrow on the right)
And text in the input can be dismissed by clicking on (x)
```

edgarordonez:
Close #194 

- Added Cypress.
- Added the test described in the issue #194 

I tried to follow these [guide](https://docs.cypress.io/guides/references/best-practices.html):


cypress/integration/User_not_logged.js
```diff
@@ -0,0 +1,23 @@
+// @noflow
+/* global cy */
+
+describe('User not logged', () => {
```
jaroslav-kubicek: the description should be something like "User can search in FAQs", filename should reflect that too, as there will be more tests related to the not logged user.

package.json
```diff
@@ -13,6 +13,7 @@
     "storybook": "start-storybook -p 9001 -c .storybook",
     "test": "jest --config=jest.json",
     "test-ci": "yarn flow:check && yarn eslint && yarn test && yarn translations:check",
+    "cypress:open": "cypress open",
```
jaroslav-kubicek: Pls add script for CI as well

dinodsaurus:
hey @edgarordonez I'm also working on cypress so you can check out #244

package.json
```diff
@@ -12,7 +12,8 @@
     "clean": "rimraf lib es dist",
     "storybook": "start-storybook -p 9001 -c .storybook",
     "test": "jest --config=jest.json",
-    "test-ci": "yarn flow:check && yarn eslint && yarn test && yarn translations:check",
+    "test-ci": "yarn flow:check && yarn eslint && yarn test && yarn translations:check && cypress run",
```
git-toni: I'm not sure this will work unless you `yarn start` it.

Actually I believe we should do integration tests on CI against the build+deployed version, not the localhost running one.

git-toni: I suggest to `yarn start` it for now and think about testing against the deployed version on another issue :slightly_smiling_face:

dinodsaurus: I agree

edgarordonez: I added `yarn start` into `test-ci` script, but this fails because are missing one dependency.

screen shot 2018-05-28 at 12 36 07

What do we do?

Issue #228, opened 2018-05-23
Upcoming / past needs to be decided by arrival time


jaroslav-kubicek:
Whether booking is future or past needs to be decided by arrival time, not by departure time.

E. g.

I flew to BCN on June 1
Today is June 2
I'm going back to PRG on June 3

===> it's still upcoming

See screen below

<img width="530" alt="screen shot 2018-05-23 at 18 45 19" src="https://user-images.githubusercontent.com/2174629/40439221-8fc9347e-5eba-11e8-9b4f-3371175aab27.png">

jaroslav-kubicek:
I might be wrong, it's listed as "Past" on Allbokings page, or maybe we need third status "Current" or something

jalbertsr:
This is a good point, "current" status for a booking that has started but not finished can be a good idea?

Also, something is not ok in that screenshot if it shows "Barcelona to Kraków and back", shouldn't it be showing something like "Kraków to Barcelona and back" ? I mean he is returning to Katowice but it's still Poland, is a rare case.

jaroslav-kubicek:
it can happen, I will ask how it should work

But definitely, it should be "Kraków to Barcelona and back"

jaroslav-kubicek:
Ok, according to MMB, it should be future, therefore expected behaviour:

- Header should be "Kraków to Barcelona and back"
- It should be "Upcoming trip" instead of "Last trip" until the final arrival back from Barcelona

jaroslav-kubicek:
<img width="546" alt="screen shot 2018-05-25 at 11 56 08" src="https://user-images.githubusercontent.com/2174629/40538698-ad6701ba-6012-11e8-80a5-9027ddbecb42.png">

The same issues apply to "All bookings" page

PR #243, opened 2018-05-25
Upcoming / past needs to be decided by arrival time

radkag:
Update header to show correctly (departure - arrival) - "Kraków to Barcelona and back"
Fix displaying "Upcoming trip" and "Last trip"


git-toni:
Please add "Closes #{issueNumber}" in the PR description. In this case should be issue #228  if I'm not mistaken? :slightly_smiling_face:

package.json
```diff
@@ -96,6 +96,7 @@
     "isomorphic-fetch": "2.2.1",
     "luxon": "1.2.0",
     "markdown-it": "^8.4.1",
+    "mockdate": "^2.0.2",
```
jaroslav-kubicek: => devDependencies

radkag: I found it weird as well, but it gave me error saying I should add it to `dependencies` instead of `devDependencies` when I tried import in test file.

Now it doesn't give error anymore.

PR #244, opened 2018-05-25
Cypress tests for navigating thru FAQs

dinodsaurus:
This PR depends on #242 , so merge that first


cypress/integration/display_faqs.js
```diff
@@ -46,6 +46,43 @@ describe('Display Static FAQ', () => {
...
+    cy.get('[data-cy=faq-article-form]').should('exist');
+
+    cy
+      .get('[data-cy=faq-article-form]')
+      .find('textarea')
+      .type(faqFeedback)
+      .should('have.value', faqFeedback);
+
+    cy
+      .get('[data-cy=faq-article-form]')
```
dinodsaurus: This actually submits form..
Im not sure if we can use fixtures with graphql?

cypress/integration/display_faqs.js
```diff
@@ -0,0 +1,101 @@
...
+  it('User can navigate into FAQ Article', () => {
+    cy.visit('/');
+    cy.get('[data-cy=btn-nonexistent-booking]').trigger('click');
+    cy.get('.signInOrBack span').contains('Sign In');
+
+    cy
+      .get('[data-cy=faq-categories]')
+      .find('a')
+      .first()
+      .click({ force: true });
```
git-toni: Why is `force` necessary?

cypress/integration/display_faqs.js
git-toni: Great point, i'm sure we can get around this by intercepting the requests while the server is running locally? 
Not sure about when testing against production tho, will look into alternatives.

cypress/integration/User_not_logged.js
```diff
@@ -0,0 +1,23 @@
+// @noflow
```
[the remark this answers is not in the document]
dinodsaurus: yea, thats from @edgarordonez PR, he solved it

cypress/integration/display_faqs.js
dinodsaurus: Because height of the `a` tag is 0. So in theory its not visible

src/StaticFAQ/FAQArticle.js
```diff
@@ -29,6 +29,7 @@ const style = css`
 
 const FAQArticle = (props: Props) => (
   <Link
+    className="faq-link"
```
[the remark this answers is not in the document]
dinodsaurus: Unfortunately `data-cy` didn't work for me on a `Link` component :(

cypress/integration/display_faqs.js
git-toni: I would comment this out for now and raise an issue linking to this obstacle. So we don't swamp the backend with feedback instances.

src/StaticFAQ/FAQArticle.js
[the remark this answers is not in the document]
git-toni: Strange, it seems to be working for me.
image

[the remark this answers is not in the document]
dinodsaurus: wow, ok will change it, thanks for testing it out!

cypress/integration/display_faqs.js
git-toni: This is interesting, if we give the `Link` a `display: block` attribute the 0 height doesn't happen and we don't have to "force" anything. 

This doesn't seem like a critical issue, but what do you think about this approach?

dinodsaurus:
@git-toni I made all requested changes

PR #255, opened 2018-05-28
Fix header on frontend

### LINKED ISSUE #253: Fix header on frontend
screen shot 2018-05-28 at 14 49 51

Header is broken because of class name collision with frontend css - `.Header` classname must be renamed to something different.

edgarordonez:
Close #253 

- Changed CSS property to solve the header broken. (I think that change the name of the class isn't needed).<br/><br/><br/><url>LiveURL: https://smartfaq-253-fix-header-frontend.surge.sh</url>

### DIFF
--- a/src/common/layout/Header.js
+++ b/src/common/layout/Header.js
@@ -23,7 +23,7 @@ const style = css`
     align-items: center;
     height: 64px;
   }
-  .Header {
+  .HeaderFAQ {
     width: 100%;
     border-bottom: 1px solid #e8edf1;
     background-color: #ffffff;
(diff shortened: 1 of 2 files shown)


src/common/layout/Header.js
```diff
@@ -47,6 +47,7 @@ const loggedInStyle = css`
   .loggedIn {
     display: flex;
     justify-content: space-between;
+    flex: 1;
```
jaroslav-kubicek: I would rather rename the class - they still can decide to change styles for `Header` anytime, which could affect us.

PR #256, opened 2018-05-28
wrong credentials redirects instead of showing error on kiwi frontend 

### LINKED ISSUE #240: Wrong credential redirects to nearest booking page
this only happens on kiwi-frontend staging, on now and gh pages it works well, probably the function that frontend passes us as a prop doesn't do a `throw error` and our try-catch can't get the error. 

instead it should show the alert system message.

gif

jalbertsr:
Close #240 <br/><br/><br/><url>LiveURL: https://smartfaq-240-fix-wrong-credentials.surge.sh</url>

### DIFF
--- a/src/KiwiLogin/__test__/KiwiLogin.test.js
+++ b/src/KiwiLogin/__test__/KiwiLogin.test.js
@@ -3,10 +3,14 @@ import * as React from 'react';
 import { shallow } from 'enzyme';
 
 import { RawKiwiLogin } from '../index';
+import { defaultToken, defaultUserId } from '../../../__mocks__/constants';
 
 describe('KiwiLogin', () => {
   const props = {
-    onLogin: async () => {},
+    onLogin: async () => ({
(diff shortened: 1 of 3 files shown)


src/KiwiLogin/__test__/KiwiLogin.test.js
```diff
@@ -35,7 +35,7 @@ describe('KiwiLogin', () => {
     const Login = wrapper.instance();
     await Login.handleSignIn(e);
     expect(e.preventDefault.mock.calls).toHaveLength(1);
-    expect(Login.props.history.push.mock.calls[0][0]).toBe('/faq/');
+    expect(Login.props.history.push.mock.calls).toHaveLength(0);
```
jaroslav-kubicek: this test shouldn't be changed, instead, `onLogin` should return object matching this flow type:

```
type Result = {|
  user_id: string,
  token: string
|}
```

PR #268, opened 2018-05-29
Enable social login to be handled by frontend

jaroslav-kubicek:


src/SignInPage/index.js
```diff
@@ -260,4 +253,5 @@ class SignIn extends React.Component<Props, State> {
   }
 }
 
-export default SignIn;
+export const RawSignIn = SignIn;
```
kantord: hmm `RawSignIn` is weird naming I think. Maybe `NativeSignIn` or `EmailSignIn` instead?

src/index.js
```diff
@@ -47,6 +48,16 @@ class Root extends React.Component<Props, State> {
     return Promise.resolve(user);
   };
 
+  onSocialLogin = async provider => {
+    const authUrl = await socialLogin(provider);
+
+    if (authUrl) {
```
jalbertsr: it shouldn't be `if(!authUrl)` ?

src/SignInPage/index.js
kantord: Maybe those should also be changed :stuck_out_tongue_winking_eye:
Like I don't feel the word "raw" describes "email" very well.

src/index.js
jaroslav-kubicek: it's still not working locally, but... 🤦‍♂️  🤦‍♂️

PR #280, opened 2018-05-30
Scenario "User can't log in SmartFAQ with bad credentials" implemented

edgarordonez:
Close #246

- This PR are using code from this other PR https://github.com/kiwicom/smart-faq/pull/279. So, I guess that we need to merge this before.

- Scenario "User can't log in SmartFAQ with bad credentials" has been implemented.<br/><br/><br/><url>LiveURL: https://smartfaq-246-cypress-user-can-not-log-bad-credentials.surge.sh</url>

PR #295, opened 2018-05-31
Added cypress fixture for submiting article question

### LINKED ISSUE #284: Find a way not to actually submit feedback form during integration tests
Currently the action of submitting the form is commented out to avoid `POST`ing many new feedback to the backend. https://github.com/kiwicom/smart-faq/blob/master/cypress/integration/LoggedOutStaticFAQ.cy.js#L75

Ideally we should test that functionality without creating a new feedback at every request.

Ideas:
- Agree with backend on some text that can be easily filtered + removed. This will allow us to test the real feedback submission.
 - Mock the response somewhow

Obstacles:
- We are running integration tests against a real deployment at surge.sh

dinodsaurus:
closes #284 
It is pretty easy when you have a single post request...
I will try the same thing on login, we will see how that goes<br/><br/><br/><url>LiveURL: https://smartfaq-cypress-comment-fixture.surge.sh</url>

### DIFF
--- a/cypress/fixtures/contact_response.json
+++ b/cypress/fixtures/contact_response.json
@@ -1,0 +1,1 @@
+{"data": {"addFAQArticleComment": {"id": "RkFRQXJ0aWNsZTo5NQ=="}}}
--- a/cypress/integration/Logged_out_static_FAQ.cy.js
+++ b/cypress/integration/Logged_out_static_FAQ.cy.js
@@ -58,6 +58,10 @@ describe('Display Static FAQ', () => {
   });
 
   it('User can submit some feedback on a FAQ article', () => {
+    cy.server();
+    cy.fixture('contact_response.json').as('contactResponse');
+    cy.route('POST', 'https://graphql.kiwi.com/', '@contactResponse');
+
     const faqFeedback = 'This is my feedback!';
     cy
       .get('.FAQArticleFeedback')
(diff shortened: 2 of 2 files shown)


dinodsaurus:
Yep, makes sense, will add it 👍 
Regarding mocking login functionality, what do you think about it, should we go in that direction or not? Please check #280

git-toni:
I believe your approach works best in the case of login, but I'm dubious we need to mock login? 

Shouldn't we integrate-test everything as close to the user as possible? Just not having side-effects such as flooding the backend with new feedback messages. Imho we can afford to *actually* login, otherwise we will need to mock bookings, etc as well.

What do you think?

dinodsaurus:
I think the same, but the only thing that concerns me is this `cy.wait(1000);` timeout.
But yea.. maybe im over engineering it at this point of time, lets see how it goes and then adapt 😁

git-toni:
One option to "clean-up" code of `cy.wait`s would probalby be overwrite some timeouts in the `cypress.json` ? https://docs.cypress.io/guides/references/configuration.html#Timeouts

edgarordonez:
Hi! @git-toni, @dinodsaurus 

I'm not sure about what it's the better approach, but probably, not has sense mock all the responses in the e2e test, anyway, I made a new command for mock all the requests, either HTTP as login or a GraphQL. You can see it in the last PR [#280](https://github.com/kiwicom/smart-faq/pull/280).

I did a mock of all the requests in that test, I guess isn't necessary, I could change it. We can discuss in that other PR. 

Thanks!

PR #301, opened 2018-06-01
fixed math for calculating body height

### LINKED ISSUE #296: Fix issues with height
Video: http://recordit.co/AenKlR3nen

**Acceptance criteria**

- SmartFAQ ALWAYS on every screen takes exactly 100% of available height and no browser scrollbar is visible
- If the content of Booking screen, FAQ article or list of FAQ categories is longer, it enables scrolling "inside" of such content

dinodsaurus:
closes #296 <br/><br/><br/><url>LiveURL: https://smartfaq-296-articles-height.surge.sh</url>


jalbertsr:
the empty scroll shouldn't appear 

screenshot

dinodsaurus:
@jalbertsr Could you please check now. Thanks 😉

jalbertsr:
I tried locally installing smart-faq on kiwi frontend and I still see the scroll bar, I think the cause might be this line or this css https://github.com/kiwicom/smart-faq/blob/master/src/App.js#L20 I can be wrong tho

dinodsaurus:
Hmmm the problem now, If we remove `overflow-y: auto;` from the `App` is that we need to take care of overflow on each component. Should I go in that direction or do you have any better suggestion? @jalbertsr

git-toni:
This is a tricky issue. If I'm not mistaken it seems that the change to make the `FAQ` section of the app a `ScrollableContent` [here](https://github.com/kiwicom/smart-faq/pull/301/files#diff-d9d9b5c037683e9e9ecf38ecc599fa1eR37) still results in a main scroll in the case for some articles(which was previously caused by the `overflow-y: scroll` at `.Body` in the same file.

If you check some articles such as `Home > Booking > Booking a flight > Can I just use part of...` or even `Home > At the airport > What is a PNR number?`(this one has clearly issues with the image, another topic).

What about setting that `ScrollableContent` mentioned above somewhere deeper excllusively?

dinodsaurus:
Thanks @git-toni , yea I will try to move each `ScrolableContent` as "deep" as posible.

jalbertsr:
works well for me! rebase plz

dinodsaurus:
@git-toni after some testing I realized that it is ok to have it inside the content page. I think it should be ok now and ready to merge 👍

PR #315, opened 2018-06-01
removed show more icon from leg

### LINKED ISSUE #307: Remove "show more" icon from leg as it's not expandable now
Remove "show more" icon on the right from leg as it's not expandable now - it's only confusing for the user.

See: https://monosnap.com/file/VGkjrD85Q2Ghin3LowLiMjSvGMMC3e

dinodsaurus:
closes #307 <br/><br/><br/><url>LiveURL: https://smartfaq-307-remove-showMore.surge.sh</url>


git-toni:
Seems we'll have to *actually* remove that for eslint not to complain.

jalbertsr:
I prefer to just comment the import we are not using here than removing the eslint rule, also maybe comment with the issue or pr link to somehow know in the future why is this commented

PR #333, opened 2018-06-04
Changed width of FAQ to 650px

### LINKED ISSUE #330: Layout columns width change
I am quite concerned with making the right column so narrow (not enough whitespace). The idea is to drive attention to the faq categories, that is why the right part is white and wider.
layout

dinodsaurus:
closes #330 
This maybe opens some responsive issues?

<br/><br/><br/><url>LiveURL: https://smartfaq-330-change-FAQ-width.surge.sh</url>

### DIFF
--- a/src/common/layout/ContentPage.js
+++ b/src/common/layout/ContentPage.js
@@ -57,7 +57,7 @@ class ContentPage extends React.Component<Props> {
             height: calc(100% - (64px));
           }
           .FAQ {
-            width: 480px;
+            width: 650px;
           }
           .BookingInfo {
             width: 548px;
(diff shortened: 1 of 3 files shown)


src/common/layout/ContentPage.js
```diff
@@ -54,7 +54,7 @@ class ContentPage extends React.Component<Props> {
             height: calc(100% - (64px));
           }
           .FAQ {
-            width: 480px;
+            width: 650px;
```
jalbertsr: the `650px` size should be **only** when logged in according to designs, when logged out it should stay as `480px`

jalbertsr:
the margins of articles are a bit different 
see: https://app.zeplin.io/project/5a8e8992554d385c76ad5aeb/screen/5aaac0e33d24e2bf8e1df3ab

PR #335, opened 2018-06-04
Set MMB booking at intropage

### LINKED ISSUE #309: Show related booking in help when in MMB
When I'm in manage my booking, looking on detail of one specific booking, SmartFAQ should be opened with that booking.

See: http://recordit.co/xjhOu7bzvf

git-toni:
Closes #309

 <br/><br/><br/><url>LiveURL: https://smartfaq-309-show-related-booking-when-mmb.surge.sh</url>

### DIFF
--- a/src/IntroPage/index.js
+++ b/src/IntroPage/index.js
@@ -12,7 +12,7 @@ import { withUser } from '../context/User';
 import type { User } from '../types';
 import FullFAQLink from '../common/FullFAQLink';
 import redirectsLoggedIn from '../common/redirectsLoggedIn';
-import { selectUrlBooking } from '../common/integration/urlProcessing';
+import { SelectUrlBooking } from '../common/integration/urlProcessing';
 
 const style = css`
   .Intro {
@@ -163,4 +163,4 @@ export class PureIntro extends React.Component<Props> {
   }
 }
 
-export default selectUrlBooking(withUser(redirectsLoggedIn(PureIntro)));
+export default withUser(redirectsLoggedIn(PureIntro));
--- a/src/IntroPage/index.js
+++ b/src/IntroPage/index.js
@@ -12,7 +12,6 @@ import { withUser } from '../context/User';
 import type { User } from '../types';
 import FullFAQLink from '../common/FullFAQLink';
 import redirectsLoggedIn from '../common/redirectsLoggedIn';
-import { SelectUrlBooking } from '../common/integration/urlProcessing';
 
 const style = css`
   .Intro {
--- a/src/IntroPage/index.js
+++ b/src/IntroPage/index.js
@@ -12,6 +12,7 @@ import { withUser } from '../context/User';
 import type { User } from '../types';
 import FullFAQLink from '../common/FullFAQLink';
 import redirectsLoggedIn from '../common/redirectsLoggedIn';
+import { selectUrlBooking } from '../common/integration/urlProcessing';
 
 const style = css`
   .Intro {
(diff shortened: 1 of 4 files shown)


src/common/integration/urlProcessing.js
```diff
@@ -0,0 +1,23 @@
...
+
+import { BookingState } from '../../context/BookingState';
+
+export const selectUrlBooking = (ComponentToWrap: React.ComponentType<any>) => {
+  const urlMatch = window.location.href.match(
+    /.*[kiwi.com|localhost:\d*]\/.*\/account\/bookings\/(\d*)$/,
+  );
+  const bookingId = urlMatch && urlMatch[1];
+  const WrappedComponent = (props: any) => (
+    <React.Fragment>
```
dinodsaurus: Why do you need a `Fragment` here?

git-toni:
Just pushed the change. It seems writing the test was a bit trickier than I expected(unless we want to mock all the Provider, which I'd rather not do). So I moved the unit test addition to another issue if that is fine with you since this feature is urgent @dinodsaurus @jaroslav-kubicek

src/common/integration/urlProcessing.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import * as React from 'react';
+
+import { BookingState } from '../../context/BookingState';
+
+export const selectUrlBooking = (ComponentToWrap: React.ComponentType<any>) => {
```
jaroslav-kubicek: I would avoid any if possible =>

`export const selectUrlBooking = <Props>(ComponentToWrap: React.ComponentType<Props>) => {`

^^ something like that should work

src/IntroPage/index.js
```diff
@@ -163,4 +164,4 @@ export class PureIntro extends React.Component<Props> {
   }
 }
 
-export default withUser(redirectsLoggedIn(PureIntro));
+export default selectUrlBooking(withUser(redirectsLoggedIn(PureIntro)));
```
jaroslav-kubicek: Have you tested this with frontend? I'm afraid it's not going to work, URL is parsed on import, not when the component is being rendered => if you access the `kiwi.com` or default MMB page and then click through app onto single booking, I expect original URL is parsed, not the one with `account/bookings/XXX`

git-toni: I was also concerned about this initially, but it seems to work well when integrated to frontend locally(although one can never be *too* sure with frontend until it's deployed to our staging branch ). I believe it might be due to lazy-loading modules, eg. mmb module is loaded once the browser route matches the specific mmb URL. [here](https://gitlab.skypicker.com/frontend/frontend/blob/master/src/mmb/browser/render.js#L31)

src/common/integration/urlProcessing.js
```diff
@@ -0,0 +1,22 @@
...
+
+const SelectUrlBooking = () => {
+  const urlMatch = window.location.href.match(
+    /.*[kiwi.com|localhost:\d*]\/.*\/account\/bookings\/(\d*)$/,
+  );
+  const bookingId = urlMatch && urlMatch[1];
+  return (
+    <BookingState.Consumer>
+      {({ onSelectBooking }) => {
+        bookingId && onSelectBooking(bookingId);
```
jaroslav-kubicek: we should still make sure it's called only once

src/IntroPage/index.js
```diff
@@ -117,6 +118,7 @@ export class PureIntro extends React.Component<Props> {
   render() {
     return (
       <div className="Intro">
+        <SelectUrlBooking />
```
jaroslav-kubicek: le'ts move it into App, it can open a different page upon opening as well

git-toni: That's true, albeit that's not even merged yet :smile: Will change :+1:

git-toni:
cc @jaroslav-kubicek Should be fixed. Together with the new `?help=xxx` param used at frontend integration.

PR #344, opened 2018-06-05
added target=blank to all ArticleDetail links

dinodsaurus:
I know this is not the best approach but I could just not get https://github.com/markdown-it/markdown-it/blob/master/docs/architecture.md#renderer to work.
But then again I dont see any potential problems with this solution, Im just inserting `target`. <br/><br/><br/><url>LiveURL: https://smartfaq-314-articleDetail-links.surge.sh</url>

### DIFF
--- a/src/common/Markdown.js
+++ b/src/common/Markdown.js
@@ -1,7 +1,8 @@
 // @flow
 import React from 'react';
+import MarkdownIt from 'markdown-it';
 
-const md = require('markdown-it')({
+const md = MarkdownIt({
   html: true,
   linkify: true,
   typographer: true,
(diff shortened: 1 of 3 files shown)


src/common/Markdown.js
```diff
@@ -11,8 +12,13 @@ type Props = {
   children: ?string,
 };
 
+const renderMarkdown = markdown => {
+  const domString = md.render(markdown);
+  return domString.replace('<a href=', '<a target="_blank" href=');
```
jaroslav-kubicek: works only for the first link in the text, check the article "Home > Booking > How do I Refer a Friend?"

dinodsaurus: Wow I totally forgot about that.. Thanks!

PR #359, opened 2018-06-06
don't allow to resize textarea

jalbertsr:
new feedback round from QA => `I am able to change size of feedback field also horizontally ` it shouldn't happen<br/><br/><br/><url>LiveURL: https://smartfaq-block-resize-textarea.surge.sh</url>


jalbertsr:
noooooo

Issue #337, opened 2018-06-05
Refactor "Manage My Booking" button to use Orbit component


kantord:
https://github.com/kiwicom/smart-faq/pull/261#discussion_r192903458

PR #369, opened 2018-06-07
Use orbit components to implement "Manage My Booking" button on mobile

kantord:
Fixes https://github.com/kiwicom/smart-faq/issues/337


src/MobileBookingHeader/MobileBookingDetail.js
```diff
@@ -54,7 +49,13 @@ const MobileBookingControls = (props: MobileBookingControlsProps) => (
       href={props.manageBookingURL}
       className="manageBookingButton"
     >
-      Manage My Booking
+      <Button
+        block
+        onClick={() => {}}
```
jaroslav-kubicek: it should probably open new tab `onClick`

kantord: I wrapped it in a link instead

jaroslav-kubicek: btw why `noopener noreferrer`?

kantord: That's a good question! :joy: Actually the code for the link was here so I just reused that. I know this is a very bad explanation given that I created the entire component.

kantord: Ah! I took it from here: https://github.com/kiwicom/smart-faq/blob/d438e98488cd0d0ff3c659c59027690ddc104076/src/SingleBookingPage/BookingDetail.js#L112

jaroslav-kubicek: it should be there, we want to track it, right? 🙂

jaroslav-kubicek: in that case let's do it in `onClick`

kantord: Like this? BTW why is the surge link broken? :thinking:

jalbertsr: your branch names are too big 😂, surge couldn't handle it

kantord: :joy:

src/MobileBookingHeader/MobileBookingDetail.js
```diff
@@ -48,14 +43,17 @@ type MobileBookingControlsProps = {|
...
-      href={props.manageBookingURL}
-      className="manageBookingButton"
-    >
-      Manage My Booking
-    </a>
+    <div className="manageBookingButton">
+      <Button
+        block
+        onClick={() => {
+          window.open(props.manageBookingURL, '_blank');
```
git-toni: Should also add `window.opener=null` :slightly_smiling_face:

git-toni: Isn't it considered an unsafe practice not to do it when `_blank`ing? One can modify the `opener` of the new tab, ie kiwi.com+smart-faq.

kantord: To my knowledge it is unsafe when it's user generated content. If someone can control the site that is opened, they'll have access to the opener `window` and do all kinds of nasty stuff. AFAIK this is not a risk in our case as the link is also on kiwi.com, so it doesn't add a new security risk.

However we can add it just in case :man_shrugging:

git-toni: ie. `window.open(props.manageBookingURL, '_blank').opener = null;`

git-toni: Agree with you it's not a risky case. I was curious as to why have it before and remove it now tho. Also it's quite cheap to add I'd say so why not? :smile:

kantord: Ok I added it :-) I chose this implementation: https://stackoverflow.com/a/46958731

PR #370, opened 2018-06-07
generate ".flow" files in lib 

### LINKED ISSUE #290: Generate ".flow" files so dummy flow-typed file is not needed
We need to generate ".flow" files in order to get rid of a flow-typed file on frontend

Acceptance criteria
- dummy file `flow-typed/npm/@kiwicom/smart-faq_vx.x.x.js` on frontend will disappear

jalbertsr:
Closes #290


jaroslav-kubicek:
does it work with frontend? that was why we removed it previously

jalbertsr:
aaah okok I'll check

PR #371, opened 2018-06-07
added skeleton animation for nearest booking

### LINKED ISSUE #361: Loading "after booking" screens
Two big animated loaders on one screen don't look very good. To prevent this, first we should fix this: #360 

Afterwards I would like to create a basic skeleton page on the left.

Design: https://zpl.io/bzM0lWG

dinodsaurus:
closes #361 <br/><br/><br/><url>LiveURL: https://smartfaq-361-booking-animation.surge.sh</url>

### DIFF
--- a/src/SingleBookingPage/NearestBooking.js
+++ b/src/SingleBookingPage/NearestBooking.js
@@ -2,6 +2,8 @@
 
 import * as React from 'react';
 import { graphql } from 'react-relay';
+import idx from "idx";
+import ContentLoader from "react-content-loader";
 
 import Loader from '../common/Loader';
 import QueryRenderer from '../relay/QueryRenderer';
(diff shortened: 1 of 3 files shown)


kantord:
wow! Amazing!!! I like it a lot. I'll take a look at the code.

src/SingleBookingPage/NearestBooking.js
```diff
@@ -27,21 +28,42 @@ const query = graphql`
...
+    if (booking) {
+      content = <BookingDetail booking={booking} />;
+    } else if (!booking) {
+      content = <BookingNotFound />;
     }
 
     if (!renderState.props) {
-      return <Loader />;
+      content = (
+        <ContentLoader
```
kantord: Code looks totally good too!

Just one idea: maybe we should create a custom component that is based on `<ContentLoader>`, but doesn't require us to specify `speed` and the colors. It's not a priority now as it's only used in one place, but once we have multiple "skeletons" it might be annoying. The same we have/had with media queries copy-pasted to every file with the same values.

git-toni:
This looks great :smile:  Why the change of style tho? ie. Moving from early `return`s to a single one with a mutable variable `content`?

src/SingleBookingPage/NearestBooking.js
jaroslav-kubicek: I would move it to separate component as well, to keep this file smaller, not issue for now

PR #372, opened 2018-06-07
Show dots only for <ul> not <ol>

### LINKED ISSUE #365: Dots should only be shown for unordered list

kantord:
fixes #365


dinodsaurus:
Ok, then we can merge it 😁

PR #392, opened 2018-06-08
GraphQL: new queries for bookings

### LINKED ISSUE #385: Refactor query for AllBookings page
Use new query for allbookings page, which returns `BookingInterface` instead of nasty `Booking` type, see: https://github.com/kiwicom/mobile/pull/729

dinodsaurus:
closes #385 <br/><br/><br/><url>LiveURL: https://smartfaq-385-bookings-query-refactor.surge.sh</url>

### DIFF
--- a/src/AllBookings/BookingCardsList.js
+++ b/src/AllBookings/BookingCardsList.js
@@ -16,6 +16,12 @@ type Props = {
   title: string,
 };
 
+type BookingVariants = {
+  BookingOneWay: React.Node,
+  BookingReturn: React.Node,
+  BookingMulticity: React.Node,
+};
+
 const BookingCardsList = (props: Props) => {
   const { booking, title } = props;
   const edges = idx(booking, _ => _.edges) || [];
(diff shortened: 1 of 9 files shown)


src/SingleBookingPage/SelectedBooking.js
```diff
@@ -23,15 +22,8 @@ type RenderState = {
...
-    booking(id: $id) {
-      type
-      oneWay {
-        ...BookingDetail_booking
-      }
-      return {
-        ...BookingDetail_booking
-      }
-      multicity {
+    node(id: $id) {
```
jaroslav-kubicek: Huh, not this query, at least now, I should have clarified it before probably more clearly - we need to keep the old `booking` because we need to support old booking ids from frontend

jaroslav-kubicek: Old booking id === number, GraphQL id === opaque string

dinodsaurus: The problem is when I refactored `BookingCardsList` `SelectedBooking` stoped working and I dunno why, because `id`s are the same 😐.. If you could checkout to `3b5fbe47163fcf651b9549389e8e248a07ac6baf` and try it that would be great!

jaroslav-kubicek: `booking(id: ID)` query accept both opaque string IDs and numerical `databaseId`, but `node` only string ID and we need both because frontend use only numerical

Therefore let's stick with `booking` query

dinodsaurus:
Switched to `databaseId` more info about the switch can be found here https://github.com/kiwicom/graphql/pull/203

src/AllBookings/BookingCardsList.js
```diff
@@ -32,26 +31,17 @@ const BookingCardsList = (props: Props) => {
         </div>
       )}
       {bookings.map(booking => {
-        const type = idx(booking, _ => _.type);
-        const id = idx(booking, _ => _.id);
-        let bookingComponent = null;
+        const id = idx(booking, _ => _.databaseId);
+        const type = idx(booking, _ => _.__typename);
+        const variants: Object = {
```
jaroslav-kubicek: `Object` is too weak type

PR #403, opened 2018-06-11
Correct booking url domain

### LINKED ISSUE #386: Show bookingUrl with correct domain to make QA happy
Clicking on "Manage my booking" button redirects to kiwi.com production page, whre you are not logged-in, which lower the user experience on staging. Let's make QA lifes better by fixing this. :heart: 

This needs to be done on GraphQL server: https://github.com/kiwicom/graphql/issues/183

kantord:
closes #386

<br/><br/><br/><url>LiveURL: https://smartfaq-correct-bookingURL-domain.surge.sh</url>

### DIFF
--- a/src/helpers/__tests__/replaceWithCurrentDomain.test.js
+++ b/src/helpers/__tests__/replaceWithCurrentDomain.test.js
@@ -1,6 +1,6 @@
 // @flow
 
-import replaceDomain from '../replaceDomain';
+import { replaceDomain } from '../replaceWithCurrentDomain';
 
 const testCases = [
   ['https://example.com', 'https://domain.net', 'https://domain.net/'],
--- a/src/helpers/__tests__/replaceWithCurrentDomain.test.js
+++ b/src/helpers/__tests__/replaceWithCurrentDomain.test.js
@@ -2,16 +2,12 @@
 
 import { replaceDomain } from '../replaceWithCurrentDomain';
 
-const testCases = [
-  ['https://example.com', 'https://domain.net', 'https://domain.net/'],
-  ['https://example.com', 'https://domain2.org', 'https://domain2.org/'],
-  ['https://kiwi.com/asdg/4/', 'https://kiwi.org', 'https://kiwi.org/asdg/4/'],
-];
+const URL = 'https://kiwi.com/asdg/4/';
+const domain = 'https://kiwi.org';
+const result = 'https://kiwi.org/asdg/4/';
 
 describe('replaceDomain', () => {
-  testCases.forEach(([URL, domain, result]) => {
-    it(`replaceDomain(${URL}, ${domain})`, () => {
-      expect(replaceDomain(URL, domain)).toEqual(result);
-    });
+  it(`replaceDomain(${URL}, ${domain})`, () => {
+    expect(replaceDomain(URL, domain)).toEqual(result);
   });
 });
--- a/src/helpers/replaceDomain.js
+++ b/src/helpers/replaceDomain.js
@@ -1,9 +1,7 @@
 // @flow
 
-import parse from 'url-parse';
-
-const replaceDomain = (URL: string, domain: string) => {
-  const parsed = parse(URL);
+const replaceDomain = (link: string, domain: string) => {
+  const parsed = new URL(link);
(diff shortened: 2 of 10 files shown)


git-toni:
Is this closing #386?

kantord:
@git-toni yes, thanks

git-toni:
I think it's simpler to merge this PR than to change the graphql sever as suggested by @jaroslav-kubicek. Agree Jarda?

kantord:
oh yes, I forgot to mention, haha. I considered changing graphql as well, but it seemed much more problematic than doing this. With graphql the base domain would have to be supplied or inferred.

A mixed solution would be to add a `directAccessPath` along `directAccessURL` on GraphQL, eliminating the need to parse the URL on the frontend

git-toni:
Agree, that was exactly what I intended to do @kantord as suggested [here](https://github.com/kiwicom/graphql/issues/183#issuecomment-396192199). And also agree that this can be a refactor later on once we really need it, so I'd go with your PR as well :slightly_smiling_face:

src/helpers/replaceDomain.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+import parse from 'url-parse';
+
+const replaceDomain = (URL: string, domain: string) => {
+  const parsed = parse(URL);
```
git-toni: what is we `const parsed = new URL(myUrlString)`? We'd save the dependency on `url-parse`.

src/helpers/replaceWithCurrentDomain.js
```diff
@@ -0,0 +1,8 @@
+// @flow
```
git-toni: Can we group both helpers in one file? They seem quite related :slightly_smiling_face:

kantord: ¡Claro! Should be cool now :wink:

kantord:
@jaroslav-kubicek let's merge this?

src/helpers/replaceWithCurrentDomain.js
```diff
@@ -0,0 +1,11 @@
...
+
+export const replaceDomain = (link: string, domain: string) => {
+  const parsed = new URL(link);
```
jaroslav-kubicek: Does it work in IE? https://developer.mozilla.org/en-US/docs/Web/API/URL#Browser_compatibility

kantord: Maybe we can polyfill it too. Or just revert to adding a dependency for this :joy: Or just using a regexp to implement a limited solution that nevertheless covers our needs.

src/helpers/__tests__/replaceWithCurrentDomain.test.js
```diff
@@ -0,0 +1,17 @@
...
+import { replaceDomain } from '../replaceWithCurrentDomain';
+
+const testCases = [
+  ['https://example.com', 'https://domain.net', 'https://domain.net/'],
+  ['https://example.com', 'https://domain2.org', 'https://domain2.org/'],
+  ['https://kiwi.com/asdg/4/', 'https://kiwi.org', 'https://kiwi.org/asdg/4/'],
+];
+
+describe('replaceDomain', () => {
+  testCases.forEach(([URL, domain, result]) => {
```
jaroslav-kubicek: This test seems to be unnecessary complicated to comprehend, just passing one arg and checking the result is fine

kantord:
> Pls also check if it’s working with older browsers, I am not sure if this is handled by babel or not

How?

src/helpers/replaceWithCurrentDomain.js
kantord: @git-toni what say you?

git-toni: Yep only Opera Mini is missing https://caniuse.com/#search=url shouldn't be an issue, should it?

git-toni:
Shouldn't `babel-preset-env` take care of browser compatibility?

kantord:
Regexp implementation pushed!

kantord:
rebased

kantord:
Tests finally green :-)

PR #432, opened 2018-06-14
Feedback validation

### LINKED ISSUE #411: Empty feedback validation
I think we should not allow for people to send us empty feedback comments.
We should add validation.

Design: https://zpl.io/2j4pw4O

dinodsaurus:
closes #411 <br/><br/><br/><url>LiveURL: https://smartfaq-feedback-validation.surge.sh</url>

### DIFF
--- a/src/StaticFAQ/ArticleFeedback/ScreenInput.js
+++ b/src/StaticFAQ/ArticleFeedback/ScreenInput.js
@@ -2,7 +2,7 @@
 
 import * as React from 'react';
 import css from 'styled-jsx/css';
-import { Heading, Text, Button, Typography } from '@kiwicom/orbit-components';
+import { Heading, Text, Button } from '@kiwicom/orbit-components';
 import { Close } from '@kiwicom/orbit-components/lib/icons';
 
 import { Box } from '../../common';
@@ -18,6 +18,11 @@ type State = {|
   error: boolean,
 |};
 
+const globalStyles = css`
+  div.inputArea p {
+    color: #d21c1c;
+  }
+`;
 const style = css`
   div.question {
     margin-bottom: 4px;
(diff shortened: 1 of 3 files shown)


src/StaticFAQ/ArticleFeedback/ScreenInput.js
```diff
@@ -96,13 +109,17 @@ class ScreenInput extends React.Component<Props, State> {
...
-          <div className="inputArea">
+          <div className={`inputArea ${error ? 'invalid' : ''}`}>
             <textarea
               data-gramm_editor="false"
               onChange={this.handleChange}
               value={this.state.comment}
-              required
             />
+            {error ? (
+              <Typography size="small" type="error">
```
jalbertsr: `Typography` is deprecated, we should use `Text` or `Heading` now

dinodsaurus: oh😓, thanks a lot for the feedback!

src/StaticFAQ/ArticleFeedback/ScreenInput.js
```diff
@@ -96,19 +115,24 @@ class ScreenInput extends React.Component<Props, State> {
...
+            {error ? (
+              <Text size="small">You haven't written any feedback.</Text>
+            ) : null}
           </div>
           <div className="button">
             <Button title="Submit" onClick={() => {}} width={100} />
           </div>
         </form>
         <style jsx>{style}</style>
+        <style jsx global>
```
git-toni: I know it's annoying not to be able to change the color of `Text` to red cause of Orbit props values :(
However I think it's better to just use a plain `<p>` tag instead of `global` styles. @jalbertsr would you agree?

jalbertsr: yes, true, agree

PR #451, opened 2018-06-19
Add Search tracking

### LINKED ISSUE #312: Add analytics measuring how many times user search in FAQ
We should measure how many times user adjust his search in order to get relevant results.

git-toni:
Closes #312 

Tracking: 
```
eventName: 'smartFAQBookingOverview', 
{ action: 'search', searchedText: *string*}'
```
This tracks every query entered in the search input, with some debounce(default 250ms).

```
eventName: `'smartFAQBookingOverview', 
{ action: 'articleClicked', queriesBeforeClick: *number*}
```

This tracks the number of searches a user made before clicking on an article. According to analytics this would indicate how long it would take a user to find a relevant result via search.<br/><br/><br/><url>LiveURL: https://smartfaq-312-track-search.surge.sh</url>


src/StaticFAQ/FAQArticle.js
```diff
@@ -26,34 +28,62 @@ const style = css`
...
+) => () => {
+  if (searchText) {
+    simpleTracker('smartFAQBookingOverview', {
+      action: 'articleClicked',
+      queriesBeforeClick,
+    });
+    resetQueriesCount();
+  }
+};
+const RawFAQArticle = (props: Props) => (
```
jaroslav-kubicek: I would split it into two components, to keep logic separated from markup

something like:
```
class TrackedFAQLink extends React.Component {
  onClick() {
    ...
  }

  render() {
   return (<FAQArticle  onClick={this.onClick} />)
  }
}
```

src/StaticFAQ/index.js
```diff
@@ -45,22 +47,36 @@ type Props = {|
...
+    logSearchQuery(e.target.value);
   };
   handleCancelSearch = () => {
     this.changeSearchText('');
+    this.resetQueriesCount();
   };
 
   renderInput = (searchText: string) => {
     const isSearching = searchText.length > 0;
+    const inputChange = e => {
```
jaroslav-kubicek: doing it in `handleSearchChange` should be fine enough, I guess there is no reason to define new function that wraps the previous one?

src/StaticFAQ/index.js
```diff
@@ -74,8 +90,15 @@ class StaticFAQ extends React.Component<Props, State> {
...
       <SearchState.Consumer>
-        {({ searchText, changeSearchText }: SearchStateType) => {
+        {({
+          searchText,
+          changeSearchText,
+          incrementQueriesCount,
+          resetQueriesCount,
+        }: SearchStateType) => {
           this.changeSearchText = changeSearchText;
+          this.resetQueriesCount = resetQueriesCount;
```
jaroslav-kubicek: this black magic is not needed if you wrap the whole component in another component that pass this as prop

git-toni: Good point, it does feel rather hacky. However if I pass those context functions as props, I will have to add them to the "raw" component's `Props` type that is receiving them, which would maybe defy the purpose of separating the tracking logic from the markup?

git-toni: I can also just create a neutral `onSearch` prop :slightly_smiling_face:

dinodsaurus: This is the worst thing about `context`, if anybody has a better solution I would love to see it 😁

git-toni: The new refactor should work around it! :smile: But your approach is good imho for necessary cases.

src/StaticFAQ/__test__/FAQArticle.test.js
```diff
@@ -0,0 +1,56 @@
+// @flow
+
+import * as React from 'react';
+import { MemoryRouter } from 'react-router';
+import { mount } from 'enzyme';
+
+//import { RawFAQArticle as FAQArticle } from '../FAQArticle';
```
jaroslav-kubicek: remove the comment?

jaroslav-kubicek: I guess filename should be updated as well? 👼

src/helpers/functionUtils.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+export const debounce = (
```
jaroslav-kubicek: small unit test would be perfect

src/StaticFAQ/index.js
jaroslav-kubicek: you had to create new components because of that, that's true, but now it's much better, I swear I can even smell the flowers & herbs while looking at it.

PR #526, opened 2018-07-06
Tablet landscape design

### LINKED ISSUE #342: Refactor idea: create react components for responsive breakpoints
Currently we use two ways to implement responsive design. One is by using `desktopOnly` or `mobileOnly` CSS classes. Another is using `react-responsive` like this:

```jsx
<MediaQuery query="screen and (min-width: 320px) and (max-width: 480px)">
  ... stuff ...
</MediaQuery>
```

I think it'd be useful to create components that encapsulate the media queries for each breakpoint, so that they don't have to be defined manually everywhere. Something like
```jsx
<MobileImplementation>
  ... stuff ...
</MobileImplementation>
```

We should also get rid of the CSS helper classes for the sake of simplicity.

### LINKED ISSUE #502: Update layout for tablets
Tablet in landscape mode should display normal desktop version, just with tighter FAQ column: https://zpl.io/a74G9qp

Tablet in portrait will be similar to mobile in landscape - intro & login screens centered  https://zpl.io/begmngq but images and texts like "We're here for you. First, let's narrow down your request." will be shown

kantord:
fixes #342 
fixes #502<br/><br/><br/><url>LiveURL: https://smartfaq-update-layout-for-tablets.surge.sh</url>

### DIFF
--- a/src/AllBookings/BookingCard/BookingCard.js
+++ b/src/AllBookings/BookingCard/BookingCard.js
@@ -16,7 +16,7 @@ import type { BookingCard_booking } from './__generated__/BookingCard_booking.gr
 
 const styles = css`
   .card {
-    width: 468px;
+    width: 100%;
     height: 128px;
     position: relative;
     border-radius: 3px;
@@ -62,7 +62,7 @@ const styles = css`
     margin-right: 8px;
     vertical-align: top;
   }
-  @media only screen and (max-width: 1180px) {
+  @media only screen and (max-width: 900px) {
     .card {
       width: 304px;
       height: 100%;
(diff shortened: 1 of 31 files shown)


jalbertsr:
for iPad I see faq with the mobile design 

not sure it should look like this :/
screenshot

kantord:
You're right. How about now?

jalbertsr:
fixed 👍 
I also noticed this in the logout/intro screens
screenshot

I think you can fix it in this issue, but it's up to you if you prefer to do it in another issue apart.

kantord:
It was pretty easy to fix it, so I just did it here.

Note: I think this MR doesn't need to be squashed :-) It might be better to not squash it even, since there's a lot of refactoring here and I think it useful to keep those in separate commits.

jalbertsr:
also for these screens pls 

gif

Issue #471, opened 2018-06-21
Cypress test depends on content in knowledgebase


kantord:
See this hotfix pull request:
https://github.com/kiwicom/smart-faq/pull/470

The cypress test in question failed because the title of the article was changed in knowledgebase. The hotfix fixes the issue, however it might be a good idea to replace this assertion with something that cannot be broken by changes outside that are not in the codebase.

dinodsaurus:
We discussed this and came to a conclusion that this is ok because seeing the faq article is the most important thing in our app and hardcoding it should be ok for now.. they are not supposed to change that often and if they do we will change it 😊

jaroslav-kubicek:
Yes, as Dino said. If that happens more often than we though, we will reopen this.

kantord:
:+1:

kantord:
@vborszeki ran into this today. 1 hour later I did, too

dinodsaurus:
ok, then just check if it exists ;)

PR #529, opened 2018-07-09
Make Cypress test more robust for anonymous user

vborszeki:
This can help a bit solving the problem discussed in issue #471


kantord:
In the commit message I changed `refactor` to `test` because it's not really a refactor

PR #530, opened 2018-07-09
hide baggage

jalbertsr:
<br/><br/><br/><url>LiveURL: https://smartfaq-hide-baggage-button.surge.sh</url>

### DIFF
--- a/src/SingleBookingPage/BookingDetail.js
+++ b/src/SingleBookingPage/BookingDetail.js
@@ -6,7 +6,7 @@ import css from 'styled-jsx/css';
 import { DateTime } from 'luxon';
 import { graphql, createFragmentContainer } from 'react-relay';
 import { withRouter } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


src/SingleBookingPage/BookingDetail.js
```diff
@@ -190,13 +190,14 @@ class BookingDetail extends React.Component<Props> {
           timeDelta && (
             <Notification hoursLeft={timeDelta} isUrgent={isUrgent} />
           )}
-        <FAQExtraInfoButton
+        {/* F.I.X.M.E Baggage is buggy */}
```
kantord: Why `F.I.X.M.E` and not `FIXME`?

jalbertsr: we have `no-warning-comments` eslint rule that don't let you write `FIXME` in comments, so I use it with `F.I.X.M.E`

kantord: :man_facepalming:

PR #534, opened 2018-07-10
mobile: fetch faq list based on booking status

### LINKED ISSUE #512: Mobile: Fetch list of FAQs based on status of the booking
Fetch list of FAQ based on the status of the booking displayed on SingleBookingPage on mobile - BEFORE_BOOKING - for not logged in user
BOOKING - for user with booking with departure in more than 48 hours
BOOKING_URGENT - for a user with booking with departure in less than 48 hours and before final arrival
PAST_BOOKING - for a user with past booking

This issue is almost identical to https://github.com/kiwicom/smart-faq/issues/324, we just need to add mobile support.

jalbertsr:
close #512 <br/><br/><br/><url>LiveURL: https://smartfaq-512-faq-list-based-on-status-mobile.surge.sh</url>

### DIFF
--- a/src/MobileBookingHeader/MobileBookingDetail.js
+++ b/src/MobileBookingHeader/MobileBookingDetail.js
@@ -6,6 +6,7 @@ import { ChevronDown, ChevronUp } from '@kiwicom/orbit-components/lib/icons';
 import { Button, TextLink } from '@kiwicom/orbit-components';
 import css from 'styled-jsx/css';
 
+import { updateFAQSection } from '../common/booking/utils';
 import bookingTypes from '../common/booking/bookingTypes';
 import { BookingState, type BookingStateType } from '../context/BookingState';
 import type { MobileBookingDetail_booking } from './__generated__/MobileBookingDetail_booking.graphql';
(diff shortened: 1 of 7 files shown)


kantord:
this is for the mobile thing only, correct?

src/MobileBookingHeader/MobileBookingDetail.js
```diff
@@ -110,6 +119,34 @@ class MobileBookingDetail extends React.Component<Props, State> {
...
+    this.updateFAQSection();
+  }
+
+  componentDidUpdate() {
+    this.updateFAQSection();
+  }
+
+  updateFAQSection = () => {
+    const { booking } = this.props;
+    const { isPastBooking } = booking;
```
kantord: hmm this is basically exactly the same as @jalbertsr's implementation, no? If so, maybe this should not be copy-pasted?

kantord: I mean the entire method + the other methods, not the specific line :-)

jalbertsr: yes I know is an ugly copy-paste, I was thinking to create a folder `help/booking` with the methods that are used by both, but since half of the `MobileBookingDetail` and related files are also a copy paste maybe we should think on somehow merge these two components into one and create them smart enough to know if mobile or desktop  should be rendered.

jalbertsr: what do you prupose? I think that since this is a bit "urgent" we can do later a `refactor` issue.

kantord: Good idea! Maybe for now it'd be enough to extract the functionality of these methods to 3 separate functions and instead of copying the entire source code then we could just call those functions.

kantord: We can refactor everything else later.

jalbertsr:
check it again @kantord

kantord:
Looks good but I'm still missing this refactor https://github.com/kiwicom/smart-faq/pull/534/commits/679be572ae538c2adab6236db1c3ca9cd0820b88#r201637862

PR #542, opened 2018-07-12
refactor folder structure of static FAQs

jaroslav-kubicek:
I couldn't help but to change it. IMHO it should clean the structure of folder a little bit 🙏 <br/><br/><br/><url>LiveURL: https://smartfaq-refactor-faq-folder-structure.surge.sh</url>


src/common/layout/ContentPage.js
```diff
@@ -7,7 +7,7 @@ import { Switch, Route } from 'react-router-dom';
 import { Desktop } from '../Responsive';
 import Header from './Header';
 import StaticFAQ from '../../StaticFAQ';
-import FAQArticleDetail from '../../StaticFAQ/FAQArticleDetail';
+import FAQArticleDetail from '../../StaticFAQ/ArticleDetail/FAQArticleDetail';
```
kantord: `ArticleDetail/FAQArticleDetail` sounds a bit too redundant. Also we have `FAQ` twice in the path.

kantord: So maybe renaming `FAQArticleDetail` would remove both redundancies

jaroslav-kubicek: fixed

src/StaticFAQ/FAQCategoryList.js
```diff
@@ -17,7 +17,7 @@ import BaggageInfo from './FAQExtraInfo/BaggageInfo';
 import BoardingPassesInfo from './FAQExtraInfo/BoardingPassesInfo';
 import FAQArticle from './FAQArticle';
 import FAQCategory from './FAQCategory';
-import Breadcrumbs from './Breadcrumbs';
+import Breadcrumbs from './breadcrumbs/Breadcrumbs';
```
kantord: Maybe use `index.js` here? :stuck_out_tongue_closed_eyes: I know we don't really like it, but this looks redundant. Normally I probably wouldn't notice this, but this is a refactoring PR.

jaroslav-kubicek: it's not possible due to relay, that's why we usually don't like `index.js` files

kantord: ah, I didn't realize that! You're right. Thanks

PR #549, opened 2018-07-16
refactor routing so router changes are propagated into url

### LINKED ISSUE #468: Refactor routing
Acceptance criteria:

- [x] Replace Memory router by custom router that will update `?help` query param on every route change, so you can always forward up-to-date link to specific article

How to test:

- as you navigate through the app, `?help` param in url should be in sync with your current position in the application, back / forward browser buttons should still work & after a full page refresh, the exact same page in the page should be loaded

jaroslav-kubicek:
After some iteration - my solution to the routing problem.

This should keep `?help` param and the current route in sync.

BREAKING CHANGE: "initialRoute" prop replaced by "route"

closes #468 <br/><br/><br/><url>LiveURL: https://smartfaq-refactor-url-routing.surge.sh</url>

### DIFF
--- a/src/helpers/QueryParamRouter.js
+++ b/src/helpers/QueryParamRouter.js
@@ -1,0 +1,86 @@
+// @flow
+
+import * as React from 'react';
+import EventListener from 'react-event-listener';
+import { createMemoryHistory as createHistory } from 'history';
+import { Router } from 'react-router';
+
+type Props = {|
+  route: string,
+  children: React.Node,
+|};
+
+class QueryParamRouter extends React.Component<Props> {
+  history: {
+    push: string => void,
+  };
+  unlisten: () => void;
+
+  constructor(props: Props) {
+    super(props);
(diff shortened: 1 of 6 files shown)


src/helpers/QueryParamRouter.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  componentWillUnmount() {
+    this.unlisten();
+  }
+
+  updateUrlParam = (pathname: string) => {
+    if (typeof window === 'undefined') {
+      return;
```
kantord: When does this happen exactly? Is it because of tests?

jaroslav-kubicek: server side rendering :)

kantord: hmm, gotcha. Maybe I'd add a comment here, or put it to a separate function.

PR #553, opened 2018-07-17
refactor: extra booking info refactored to avoid gotcha with context

jaroslav-kubicek:
- refactoring to avoid gotcha with React context - described here: https://reactjs.org/docs/context.html#caveats<br/><br/><br/><url>LiveURL: https://smartfaq-refactoring-extra-info.surge.sh</url>

### DIFF
--- a/src/StaticFAQ/FAQExtraInfo/BoardingPassesInfo/index.js
+++ b/src/StaticFAQ/FAQExtraInfo/BoardingPassesInfo/index.js
@@ -105,13 +105,11 @@ class BoardingPassesInfo extends React.Component<Props> {
             <QueryRenderer
               query={nearestInfoBoardingPasses}
               render={this.renderBoardingPassesCard}
-              cacheConfig={{ force: true }}
             />
           ) : (
             <QueryRenderer
               query={selectedInfoBoardingPasses}
               render={this.renderBoardingPassesCard}
-              cacheConfig={{ force: true }}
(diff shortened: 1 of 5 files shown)


src/StaticFAQ/FAQExtraInfo/BoardingPassesInfo/index.js
```diff
@@ -105,13 +105,11 @@ class BoardingPassesInfo extends React.Component<Props> {
             <QueryRenderer
               query={nearestInfoBoardingPasses}
               render={this.renderBoardingPassesCard}
-              cacheConfig={{ force: true }}
```
jaroslav-kubicek: I had idea baggage info doesn't need to be refetched on every toggle 🤔

kantord: What do you mean? I think if you delete this line, it will indeed be re-fetched every time, no?

jaroslav-kubicek: nope, it uses cache by default and `force: true` will enforce relay to call GraphQL

kantord: Ah I see

kantord:
Then just one last note. If this fixes a behavior, I think it should be "fix" in the commit message

PR #562, opened 2018-07-19
fix: disable background scroll & show previously displayed booking

### LINKED ISSUE #380: Remove background scrolling
https://monosnap.com/file/TJEUECGJXSSLAczD02NRtyCRoSI6J0

### LINKED ISSUE #383: Show previously displayed booking
Acceptance criteria

- When I open SmartFAQ anywhere on the frontend, it shows the nearest booking
- When I open SmartFAQ in MMB in booking detail page, this specific booking is opened
- When I open SmartFAQ anywhere, MMB booking detail included, and select different booking from the allbookings page, then I close SmartFAQ and reopen it again no matter on which page of the frontend, SmartFAQ is smart enough to display booking previously selected by a user 

=> in other words, if the user intentionally selected one booking from the list, keep track of such selection so it opens the same booking later again, but don't implement such memoization for specific booking which was displayed just because the user was on MMB detail page, makes sense?

jaroslav-kubicek:
fix #380, fix #383

- disables background scroll
- removes `locale` prop
- enables SmartFAQ to be always rendered with new `isOpen`=true/false prop, so contexts are preserved => previously displayed booking is displayed

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -3,10 +3,12 @@
 
 import * as React from 'react';
 import css from 'styled-jsx/css';
+import classNames from 'classnames';
 import { I18nextProvider } from 'react-i18next';
 import 'url-search-params-polyfill';
 
 import initTranslation from './initTranslation';
+import EnLocale from '../i18n/en/translation.json';
 import Routes from './Routes';
 import { CloseContext } from './context/Close';
 import { LanguageContext } from './context/Language';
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -138,6 +152,9 @@ class App extends React.PureComponent<Props, State> {
         <style jsx global>
           {style}
         </style>
+        <style jsx global>
+          {isOpen ? 'body { overflow-y: hidden; }' : ''}
```
kantord: hmm, does this work in Safari?

jaroslav-kubicek: seems like it works 🤔

jaroslav-kubicek:
CI fixed & rebased, cc @kantord

PR #591, opened 2018-07-30
feat(analytics): analytics for login

### LINKED ISSUE #464: Add login type to analytics events
Ideally we should send login info together with our tracking events. Example with the main event `smartFAQ` with clickOnHelp/close actions

```
eventName: "smartFAQ"
attributes: {
  action: "clickOnHelp"
  login: null / magicLink / google / FB / kiwiAccount
}
```
Not sure if we can provide all that info honestly, we should explore it. 
Obviously not the magicLink since it's not even implemented yet :smile:

jalbertsr:
Close #464 <br/><br/><br/><url>LiveURL: https://smartfaq-464-login-analytics.surge.sh</url>

### DIFF
--- a/src/KiwiLogin/index.js
+++ b/src/KiwiLogin/index.js
@@ -12,6 +12,7 @@ import BackButton from '../common/buttons/BackButton';
 import Input from '../common/Input';
 import { withLogin } from '../context/User';
 import image from '../../static/woman-with-laptop@2x.jpg';
+import { simpleTracker } from '../helpers/analytics/trackers';
 import type { onLogin } from '../types';
 
 const style = css`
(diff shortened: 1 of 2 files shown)


src/SignInPage/ResponsiveSocialButton.js
```diff
@@ -15,6 +16,12 @@ type Props = {
   children: string,
 };
 
+const clickOnSocialLoginTracker = (type: string) =>
```
kantord: Why is this called `clickOnSocialLoginTracker`? Sounds as if it was clicking on something

src/KiwiLogin/index.js
```diff
@@ -155,6 +156,12 @@ type State = {|
   isLoading: boolean,
 |};
 
+const clickOnKiwiLoginTracker = () =>
```
kantord: same as here https://github.com/kiwicom/smart-faq/pull/591/files#r206534413

src/SignInPage/ResponsiveSocialButton.js
jalbertsr: I wanted to use similar name as the event tracker name, but yes, you're right, removing the `clickOn` is better for the function name. fixed :)

Issue #563, opened 2018-07-19
Add cypress test to verify Boarding passes


jaroslav-kubicek:
Add cypress tests that will verify booking pass functionality

test these things
- booking passes info is displayed when clicking on the button
- a specific FAQ category is shown
- when clicking on "Home", the previous root "section" category is shown (e.g. remember what was the first FAQ in the list before clicking on the button and compare that it's same after clicking on "Home")

as the best practice, don't write everything in one `it('...')` test step

PR #597, opened 2018-07-31
Add missing integration tests for boarding passes info

vborszeki:
mentioned in #563
<br/><br/><br/><url>LiveURL: https://smartfaq-add-boarding-passes-integration-test.surge.sh</url>

### DIFF
--- a/cypress/integration/BoardingPassesInfo.js
+++ b/cypress/integration/BoardingPassesInfo.js
@@ -1,0 +1,66 @@
+// @noflow
+/* global
+  Cypress,
+  cy
+*/
+
+describe('Boarding Passes Info', () => {
+  beforeEach(() => {
+    cy.visit('/');
+    cy.get('[data-cy=btn-existent-booking]').click();
+    cy.get('[data-cy=link-kiwi-login]').click();
+
+    cy.get('[data-cy=input-email]').type(Cypress.env('TEST_USER_EMAIL'));
+    cy.get('[data-cy=input-password]').type(Cypress.env('TEST_USER_PASSWORD'));
+    cy.get('[data-cy=btn-sign-in]').click();
+  });
+
+  it('should display boarding passes info', () => {
+    cy.get('[data-cy=btn-boarding-passes]').click();
+
+    cy
+      .get('.boardingPassesCard')
+      .find('p')
+      .contains(
+        'Here, you can download your boarding passes or see when they will become available.',
+      )
+      .get('.boardingPassesRow')
+      .should('exist');
+  });
+
+  it('should show a specific FAQ', () => {
+    cy
+      .get('[data-cy=btn-boarding-passes]')
+      .click()
+      .get('[data-cy=faq-article-link]')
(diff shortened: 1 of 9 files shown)


cypress/integration/BoardingPassesInfo.js
```diff
@@ -0,0 +1,67 @@
...
+/* global
+  Cypress,
+  cy
+*/
+/* eslint-disable jest/valid-expect */
+
+describe('Boarding Passes Info', () => {
+  beforeEach(() => {
+    cy.visit('/');
+    cy.get('[data-cy=btn-existent-booking]').click();
```
jaroslav-kubicek: this looks same as in `UserCanLogIntoSmartFAQ.js`, do you think you could create a custom command for it? 🙂  You can take inspiration from FE too, e.g. their `mmbLoad`

cypress/integration/BoardingPassesInfo.js
```diff
@@ -0,0 +1,67 @@
+// @noflow
+/* global
+  Cypress,
+  cy
+*/
+/* eslint-disable jest/valid-expect */
```
jaroslav-kubicek: we might need to set this for the whole cypress subfolder 🤔

cypress/integration/BoardingPassesInfo.js
```diff
@@ -0,0 +1,67 @@
...
+    cy
+      .get('[data-cy=btn-boarding-passes]')
+      .click()
+      .get('[data-cy=faq-article-link]')
+      .contains('PNR numbers');
+  });
+
+  it('should show the previous root category after clicking home', () => {
+    cy
+      .wait(2000)
```
jaroslav-kubicek: why wait? Cypress waits by default, so usually it's not needed

cypress/integration/BoardingPassesInfo.js
```diff
@@ -0,0 +1,67 @@
...
+      .get('[data-cy=faq-categories]')
+      .find('a')
+      .first()
+      .then($faqCard => {
+        cy
+          .get('[data-cy=btn-boarding-passes]')
+          .click()
+          .get('.boardingPassesCard');
```
jaroslav-kubicek: IMHO this `get()` has no function here, right?

vborszeki: that's right, I have used it just to make sure, that after the click the view has changed, but the presence of the home breadcrumb is also indicating this change, so I've removed it 🙂

cypress/integration/BoardingPassesInfo.js
vborszeki: It was because of a very strange behaviour, without wait the test was breaking locally. I was also checking the perex of the FAQ category, and there was a difference in a comma between the expected and the actual value. In the initial render and after clicking home, we are querying the GraphQL server with different section variables, and for some reason for the same category the perex differs in one character. However it's not totally clear for me, why the wait solved that. 
🤔🤯 Now I'm comparing just the titles there.

PR #601, opened 2018-07-31
Update build step

jaroslav-kubicek:
Changes
- ship also unprocessed ES modules => smaller bundle size on FE
- flow type defs for `react-i18next` added
- introduce `flowtype/type-import-style` - mixing import of modules with import of types breaks the usage of `transform-flow-comments` in ES modules

yes:
```
import { translate } from 'react-i18next';
import type { TFunction } from 'react-i18next';
```

no:
```
import { translate, type TFunction  } from 'react-i18next';
```

because it would produce broken code: `import { translate, /* type TFunction */  } from 'react-i18next';`<br/><br/><br/><url>LiveURL: https://smartfaq-update-build-step.surge.sh</url>


package.json
```diff
@@ -32,6 +33,7 @@
     "commitmsg": "commitlint -E GIT_PARAMS"
   },
   "main": "lib/App.js",
+  "module": "es/App.js",
```
kantord: why this?

kantord: I mean how does this work

jalbertsr: related maybe to https://github.com/kiwicom/smart-faq/pull/601/files#diff-e56633f72ecc521128b3db6586074d2cR42

jaroslav-kubicek: that's de-facto standard for shipping code with ES2015 modules, if the consumer understand it, this will be used, otherwise `main`

jaroslav-kubicek: see this proposal: https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md

jaroslav-kubicek: build for client with webpack understands it => will use `module`
e.g. node.js in SSR which  doesn't support ES2015 modules => will use `main`

PR #604, opened 2018-08-01
fix: disable background scroll on mobile safari

jaroslav-kubicek:
Related to #380 <br/><br/><br/><url>LiveURL: https://smartfaq-fix-mobile-safari-scroll.surge.sh</url>

### DIFF
--- a/src/helpers/MobileSafariScroll.js
+++ b/src/helpers/MobileSafariScroll.js
@@ -1,0 +1,28 @@
+// @flow
+
+import * as React from 'react';
+
+import { Mobile } from '../common/Responsive';
+
+type Props = {|
+  isOpen: boolean,
+|};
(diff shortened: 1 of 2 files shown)


src/helpers/MobileSafariScroll.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+
```
kantord: I miss some documentation here, because without context the purpose of this component is unclear.
I think a one-sentence summary + link to GitHub/Slack conversations about this should be enough

jaroslav-kubicek: fixed

kantord: I see it now! Thanks

Issue #474, opened 2018-06-21
Use semantic-release to automate the npm publication process


kantord:
see https://skypicker.slack.com/archives/C9K87QN6Q/p1529590764000002

kantord:
@jaroslav-kubicek plz https://glebbahmutov.com/blog/how-to-setup-semantic-release-on-circle-ci/

I believe I don't have authorization to do that alone :-)

jaroslav-kubicek:
> I believe I don't have authorization to do that alone :-)

you can ask on slack for membership in kiwicom github organization 🙂

PR #607, opened 2018-08-01
fix issues with semantic release

kantord:
https://github.com/kiwicom/smart-faq/issues/474<br/><br/><br/><url>LiveURL: https://smartfaq-fix-issues-with-semantic-release.surge.sh</url>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ version: 2
 defaults: &defaults
   working_directory: ~/smart-faq
   docker:
-    - image: circleci/node:8.0
+    - image: circleci/node:9.0
 jobs:
   test-ci:
     <<: *defaults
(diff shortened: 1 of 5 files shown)


jalbertsr:
maybe you should update DOCs on how it would work now, last time commitlint failed https://circleci.com/gh/kiwicom/smart-faq/4995

kantord:
@jalbertsr I added docs here https://github.com/kiwicom/smart-faq/pull/607/commits/8c8817daa4ee9af0d6e26ae7218df7fd86cb7bab

Do you think some more information is needed?

kantord:
fixed the commitlint issues too

kantord:
@jaroslav-kubicek no, actually GitHub releases with release notes should be created semantic-release.

For example here all of my release notes were generated by semantic-release:
https://github.com/kantord/just-dashboard/releases/tag/v0.8.0

jaroslav-kubicek:
cool, that looks great, the no. of releases will just increase, so we will have usually multiple release notes like this for one MR, right? it also means we should disable squashing after this is merge...

kantord:
> we will have usually multiple release notes like this for one MR, right?

One CI run will always generate one release notes. Each line in the release notes corresponds to one commit. Which means that most of the time one Pull Request will produce one release, since we usually merge to master directly. Essentially, we'll release everything immediately after merge.

> it also means we should disable squashing after this is merge...

Maybe, but not necessarily. I mean we should still use squash as a method of producing more user-friendly commit messages. For example if we have "feat: implement card title" and "feat: implement card perex" we could squash them into "feat: implement card". The only thing we **have to** watch out for is that the squashed commits have to obey the format as well, and have to be user-friendly. To avoid mistakes with that, commits will be validated after commit, and also before release.

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ version: 2
 defaults: &defaults
   working_directory: ~/smart-faq
   docker:
-    - image: circleci/node:8.0
+    - image: circleci/node:9.0
```
jalbertsr: why this change? is it strictly necessary? I think the LTS version is `8.11.3`

kantord:
rebased @jalbertsr @jaroslav-kubicek

kantord:
rebased again @jalbertsr :roll_eyes:

kantord:
Merge?

PR #618, opened 2018-08-06
simple token for mobile

### LINKED ISSUE #611: "Select another booking" doesn't work in mobile layout with simple token
When I open MMB with a simple token on mobile, then click on "Select another booking", it's not working properly.

Please, implement one component for this button with all necessary logic to not duplicate the code in `bookingItem/Header.js` and `MobileBookingDetail.js`

jalbertsr:
Close #611 <br/><br/><br/><url>LiveURL: https://smartfaq-611-simple-token-for-mobile.surge.sh</url>

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1,5 +1,5 @@
 # source: https://graphql.kiwi.com/
-# timestamp: Thu Jul 26 2018 10:51:32 GMT+0200 (Central European Summer Time)
+# timestamp: Fri Aug 03 2018 15:54:40 GMT+0200 (CEST)
 
 schema {
   query: RootQuery
@@ -214,6 +214,7 @@ type Booking {
   availableWhitelabeledServices: WhitelabeledServices
   onlineCheckinIsAvailable: Boolean
   insurancePrices: [InsurancePrice]
+  authToken: String
   oneWay: BookingOneWay
   return: BookingReturn
   multicity: BookingMulticity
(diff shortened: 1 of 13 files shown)


src/MobileBookingHeader/MobileUserDetail.js
```diff
@@ -4,8 +4,9 @@ import * as React from 'react';
 import css from 'styled-jsx/css';
 import { Text } from '@kiwicom/orbit-components';
 
-import { UserContext, type UserContextType } from '../context/User';
+import { UserContext } from '../context/User';
 import SignOutButton from '../common/layout/SignOutButton';
+import type { UserContextType } from '../context/User';
 
 const MobileUserDetailStyle = css`
```
jalbertsr: For the user detail section of the mobile header when the user uses simple token we don't have the name of the user account, we only have the names of the passengers in the booking.

What should we show instead? 

full login:
screenshot

simple token login: 
screenshot

@jaroslav-kubicek

kantord: How about hiding the entire tab altogether? Does the logout button have a purpose?

kantord: From the two options I prefer the second one.

jalbertsr: good point! you shouldn't be available to sign out, because you're not fully logged in

kantord: Well, just make sure also that it's not possible to compromise user data. What if I log in on a publicly used device? Will an unauthorized person be able to download my boarding passes/tickets and see my personal data on them?

kantord: Maybe it's FE's responsibility though

PR #622, opened 2018-08-07
test(smartfaq): add test for SSR

### LINKED ISSUE #556: Add test verifying Smart FAQ is rendered without error in SSR
... to make sure we have this possibility to keep a chance to replace https://www.kiwi.com/en/helpcenter/ if it's requested.

jalbertsr:
close #556 <br/><br/><br/><url>LiveURL: https://smartfaq-556-test-smartfaq-ssr.surge.sh</url>


kantord:
To test this I would actually set up server-side rendering and run Cypress tests. This test tests a functionality that doesn't exist and is only defined in the test itself. I don't know if it would be possible to test server-side rendering without having server-side rendering. And also it would be more useful to verify that the functionality of our site works with server-side rendering, and it seems logical to use Cypress tests to do that.

jaroslav-kubicek:
Do you see any specific concerns @kantord ? From my perspective, it should be enough

jalbertsr:
I see the point on what @kantord says, but I think is overkill running cypress test 2 times(the ones we have now + for SSR). It's true that this test doesn't test functionality, but it tests that we have the possibility to use SSR and it would render correctly if we need to use it in a future, which I think is enough.

kantord:
I know that our CI is already slow, but we could run these tests in parallel with the normal Cypress tests, and once we make a final decision to use SSR, we can just remove the Cypress tests that don't use the SSR. (Or the same if we decide to not use SSR)

kantord:
next.js should be almost zero config

jaroslav-kubicek:
I still believe it's enough for now. FE actually uses the same + supertest and express too.

kantord:
I'm still not convinced that these tests are useful to us.

jaroslav-kubicek:
> I'm still not convinced that these tests are useful to us.

huh, ok, would you accept if we merge this one, since something is better than nothing and then create follow up issue in which we will need to deal with SSR+relay modern, a discussion here: https://github.com/facebook/relay/issues/1881

Cypress tests for this are too heavyweight IMHO.

kantord:
pls rebase

kantord:
I still don't see the usefulness of this, however, I also don't think that it's important enough to spend a lot more time arguing about this, since I don't see a huge harm done either. So I dismissed my review

PR #624, opened 2018-08-07
Fix bad commit message

kantord:


jaroslav-kubicek:
I changed master already 👼

PR #625, opened 2018-08-07
feat(app): trap focus inside app when it's open

kantord:
<br/><br/><br/><url>LiveURL: https://smartfaq-add-focus-trap-to-app.surge.sh</url>


kantord:
@jaroslav-kubicek you are right. I remember now. That's why I did it like this: https://gitlab.skypicker.com/frontend/frontend/commits/trap-tab-key-within-sibebar how about rebasing and merging this?

Issue #396, opened 2018-06-11
Use <UserStatus.LoggedIn> everywhere, and add <UserStatus.LoggedOut>


kantord:

jalbertsr:
closed by #630

PR #630, opened 2018-08-08
Add searchbar to header

### LINKED ISSUE #558: Search in header
According to our data, most of our users start by clicking on categories and do not search. However, they do search when clicking through categories was not successful for them.

To help them with this use case we will make search smaller and put it into the header (so there is a bigger focus on tiles in the left panel) and make it available also when they are in an article. 

At the same time, we will get rid of the help centre link.

Zeplin (it is in the screen with emergencies): https://zpl.io/V1qE7ZJ

vborszeki:
closes #558
<br/><br/><br/><url>LiveURL: https://smartfaq-search-in-header.surge.sh</url>


jalbertsr:
It breaks search input for mobile:
screenshot

For iPad I think the search input should stay in the right part, becoming progressively smaller.
screenshot

kantord:
as @jalbertsr says, it breaks the input field on mobile. However, only if you are logged in. When you are not logged in, it's fine:
2018-08-09-160529_761x1089_scrot

jalbertsr:
This also closes https://github.com/kiwicom/smart-faq/issues/396 👍 I've checked and we're using `<UserStatus.LoggedIn>` everywhere we can.

jaroslav-kubicek:
thanks all for shipping this

> It breaks search input for mobile

let's add cypress test for search on mobile, so we have covered our back here.

vborszeki:
> let's add cypress test for search on mobile, so we have cover our back here.

@jaroslav-kubicek sure, I've added #640

kantord:
Search in mobile still broken

vborszeki:
@kantord the fix is on it's way 🚂

vborszeki:
@jalbertsr I have asked @mjancik about the specific cases, for now I would leave the layout as it is.

PR #644, opened 2018-08-10
Fix boarding passes integration test

vborszeki:
I had to add wait to this test, because FAQs are rendered earlier then bookings and after bookings are fetched, FAQs related to them are displayed. It wasn't breaking before, because the FAQs corresponding to the upcoming trip were the same as in the initial render, but there are cases when it can brake (close departure date for trip and displaying urgent contact). Maybe we could also solve the root cause (try out React Suspense later 🙃) and avoid loading irrelevant FAQs.


kantord:
Pls fix commit message. It should be "test". Because this is not adding a new feature for users

PR #679, opened 2018-08-23
fix(mobile-back-button): fix back button on mobile

kantord:
see https://gitlab.skypicker.com/frontend/frontend/merge_requests/5753#note_746276


MagdalenaAlmanova:
I think, that it is ok... Please, for future don't put description during skypicker. I haven't an access here. Thanks :)

PR #700, opened 2018-08-29
fix: avoid sending unwanted GraphQL booking query when opening SmartFAQ

### LINKED ISSUE #661: Fix failing GraphQL queries
As @mrtnzlml says:

> There are thousands of your queries failed and it’s happening for a long time already [..]. For example `NearestBookingQuery` failed 1594x in last 24 hours.

jaroslav-kubicek:
fixes #661

The problem was that NearestBookingQuery was sent before url with specific bookingId was processed<br/><br/><br/><url>LiveURL: https://smartfaq-fix-sending-unwanted-graphql-query-ii.surge.sh</url>


src/context/SelectedBooking.js
```diff
@@ -4,11 +4,14 @@
...
 
 type Props = {|
+  // isOpen required here only to trigger processUrl when opening/closing SmartFAQ
+  isOpen: boolean, // eslint-disable-line react/no-unused-prop-types
   children: React.Node,
 |};
 type StateValues = {
   selectedBooking: ?number,
   bookingPage: 'SINGLE_BOOKING' | 'ALL_BOOKINGS',
+  userSelectedBooking: boolean,
```
kantord: Why is this required? Isn't `selectedBooking` `null` when and only when `userSelectedBooking` is `false`?

kantord:
> The problem was that NearestBookingQuery was sent before url with specific bookingId was processed

This only happened when people people opened a single booking using a link, correct?

src/context/SelectedBooking.js
jaroslav-kubicek: I put there explanation into a comment, to keep our initial behaviour, you need to differentiate when specific booking was set automatically by url and when user specifically went to list of another bookings and selected one - in such case I don't want to change his settings

kantord: :+1:

PR #711, opened 2018-09-04
fix(warnings): fix warnings

### LINKED ISSUE #656: Get rid of lint warnings
Probably there's no reason why we should have these warnings:

2018-08-14-152401_1105x495_scrot

I believe most of them can be fixed, the one about `dangerouslySetInnerHTML` should be ignored.

jalbertsr:
close #656


src/StaticFAQ/breadcrumbs/__test__/CustomBreadcrumbs.test.js
```diff
@@ -14,42 +14,16 @@ describe('CustomBreadcrumbs', () => {
...
-      </WithRouter>,
-    );
-    const breadcbumbsElements = wrapper.find('.breadcrumb');
-    // 6 because there are 2 for mobile 2 for desktop
-    expect(breadcbumbsElements).toHaveLength(2);
-    expect(breadcbumbsElements.first().text()).toBe('Home');
-    expect(breadcbumbsElements.last().text()).toBe('Article');
-  });
-  // make it work with react-responsive
-  xit('should render shorter breadcrumbs', () => {
```
jalbertsr: this is old, we don't use `...` in breadcrumbs any more, now is the same for desktop and mobile.

jalbertsr: we show `Article` instead of the article name.

PR #713, opened 2018-09-04
fix(booking-types): rename booking-types - 267

### LINKED ISSUE #267: We have a component named `<Return />`
In `bookingTypes` we have
-  Return
-  Multicity
- Oneway

`<Return />` is a bit confusing, so it should be renamed to something like `<ReturnTrip />`. For consistency's sake, all the other component names in `bookingTypes` should be suffixed with Trip as well (`MultiCityTrip` and `OneWayTrip`)

jalbertsr:
close #267<br/><br/><br/><url>LiveURL: https://smartfaq-267-rename-files.surge.sh</url>

### DIFF
--- a/src/SingleBookingPage/bookingTypes/MulticityOverlayTrip.js
+++ b/src/SingleBookingPage/bookingTypes/MulticityOverlayTrip.js
@@ -3,7 +3,7 @@
 import * as React from 'react';
 import { createFragmentContainer, graphql } from 'react-relay';
 
-import Multicity from './Multicity';
+import MulticityTrip from './MulticityTrip';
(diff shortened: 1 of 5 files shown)


src/SingleBookingPage/bookingTypes/MulticityOverlayTrip.js
```diff
@@ -3,7 +3,7 @@
 import * as React from 'react';
 import { createFragmentContainer, graphql } from 'react-relay';
 
-import Multicity from './Multicity';
+import Multicity from './MulticityTrip';
```
vborszeki: I think you forgot to rename the import itself and it's usages below 🙂

jalbertsr: good catch, fixed

PR #750, opened 2018-09-20
Add translations nitrolib

jalbertsr:
<br/><br/><br/><url>LiveURL: https://smartfaq-add-translations-nitrolib.surge.sh</url>


jalbertsr:
@kantord @vborszeki @jaroslav-kubicek some little things may change, but you can start CR

.flowconfig
```diff
@@ -15,6 +15,7 @@ flow-typed
 
 [options]
 emoji=true
+module.ignore_non_literal_requires=true
```
jaroslav-kubicek: is it possible to put there some comment why?

src/StaticFAQ/FAQCategoryList.js
```diff
@@ -280,14 +284,16 @@ class RawFAQCategoryList extends React.Component<Props> {
...
+    const { categoryId, section, showGuaranteeArticle, language } = this.props;
+    const locale = fromLanguageToLocale(language);
 
     if (categoryId) {
       return (
         <QueryRenderer
           query={querySubcategory}
           render={this.renderSubcategory}
           variables={{ id: categoryId }}
+          createEnvironment={createEnvironment(null, locale)}
```
jaroslav-kubicek: our `QueryRenderer` doesn't accept such property, an environment is created inside of it automatically

README.md
```diff
@@ -60,9 +60,45 @@ the squashed commit also according to the convention.
...
 
-- Check out [react-i18next](https://react.i18next.com/) and [i18next](https://www.i18next.com/) docs for more info.
+- Check out [nitrolib](https://github.com/kiwicom/nitrolib) docs for more info.
 - [DOCs for Phraseapp](https://phraseapp.com/docs/) - where the translations are stored for translators
-- List of all supported languages is defined by file `i18n/languages.js`
+- List of all supported languages is defined by file `src/translations/langInfos.js`
+
+### How to use translations in code
+
+- Use `Text` from [nitrolib](https://github.com/kiwicom/nitrolib) as `Trans` and don't confuse it with `Text` from [Orbit](https://github.com/kiwicom/orbit-components):
```
kantord: wow, nitro and orbit should co-operate to merge this into one component with optionally disable-able translations :-D

README.md
```diff
@@ -60,9 +60,46 @@ the squashed commit also according to the convention.
...
+  values?: Values, // a map of placeholders and their values
+  html?: boolean, // make it 'true' if your translation string contains HTML elements
+};
+```
+
+Example:
+```js
+const Submit = () => (
+  <Button>
+    <Trans t={__("smartfaq.submit_button")} />
```
kantord: I don't really like this TBH because it makes it harder to find things using full text search. For example how would I find all components that contain a specific word, say "Integration"? Now it's very easy, for example `grep -i "integration" -r src`.

After this change we will have to do a full text search on all translations, and then a full text search for each translation key the translation of which contains the specific word. That makes it quite hard to make project wide changes such as changing "bags" to "baggage" in every component.

.flowconfig
```diff
@@ -16,5 +16,8 @@ flow-typed
 [options]
 emoji=true
 
+# stop flow complaining when you use require() with something other than a string literal.
```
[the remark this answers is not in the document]
jalbertsr: Ok, I changed it to an inline `// $FlowExpectedError` instead of global disabling it.

PR #761, opened 2018-09-25
Add Cypress tests for back button mobile

### LINKED ISSUE #755: Add Cypress tests to verify correct behaviour of back button on mobile
When the user is logged in and
- [ ] in an FAQ category, it should navigate to the root categories
- [ ] in an FAQ article, it should go to the category to which the article belongs
- [ ] in an FAQ article displayed from search, it should return to search

vborszeki:
closes #755<br/><br/><br/><url>LiveURL: https://smartfaq-cypress-test-for-back-button-mobile.surge.sh</url>

### DIFF
--- a/cypress/integration/BackButtonMobile.js
+++ b/cypress/integration/BackButtonMobile.js
@@ -1,0 +1,58 @@
+// @noflow
+/* global
+  Cypress,
+  cy
+*/
+
+describe('Back button mobile', () => {
+  beforeEach(() => {
+    cy.viewport('iphone-6');
+    cy.signIntoAccount();
+    cy.wait(2000);
+  });
+
+  it('should navigate to root categories, when clicking back on faq category page', () => {
+    cy
+      .get('[data-cy=faq-categories]')
+      .get('.faq-category')
+      .first()
+      .then($initialRootCategory => {
+        $initialRootCategory.click();
+
+        cy.reload();
+        cy.wait(2000);
+        cy.get('[data-cy=back-button-mobile]').click();
+
+        cy
+          .get('.faq-category')
+          .first()
+          .then($rootCategory => {
+            expect($initialRootCategory.text()).to.eq($rootCategory.text());
+          });
+      });
+  });
+
+  it('should navigate to faq category, when clicking back on faq article page', () => {
+    cy.getFirstFaqCategory().click();
+
+    cy.get('[data-cy=faq-article-link]').click();
+    cy.reload();
+    cy.get('[data-cy=back-button-mobile]').click();
+
+    cy.get('[data-cy=faq-article-link]').should('exist');
+  });
+
+  it('should navigate to search, when clicking back on faq article displayed from search page', () => {
+    cy.get('[data-cy=search-button]').click();
+    cy.get('[data-cy=input-staticFAQ]').type('sports equipment');
+
+    cy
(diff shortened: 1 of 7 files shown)


cypress/integration/BackButtonMobile.js
```diff
@@ -0,0 +1,58 @@
...
+/* global
+  Cypress,
+  cy
+*/
+
+describe('Back button mobile', () => {
+  beforeEach(() => {
+    cy.viewport('iphone-6');
+    cy.signIntoAccount();
+    cy.wait(2000);
```
vborszeki: the reason why _wait_ is needed here, is because _back button for mobile_ is used just for logged in user, and for logged in user the root FAQs can change depending on status of booking, but initially the general root FAQ list is rendered

jaroslav-kubicek: then pls add explanation comment so this info is preserved in code

jaroslav-kubicek: or to avoid wait, change it to wait for FAQ to be loaded and then wait for "about Kiwi.com" category to disappear which means some after booking FAQ section category was loaded. I would expect 2 sec wait to be quite flaky.

vborszeki:
@jaroslav-kubicek changes done

cypress/support/commands/commonCommands.js
```diff
@@ -55,3 +54,10 @@ Cypress.Commands.add('loadFAQArticle', () => {
     .first()
     .click();
 });
+
+Cypress.Commands.add('waitForFaqsBasedOnBookingStatusToLoad', () => {
+  cy
```
jaroslav-kubicek: one more thing - should this command also verify that _some_ faqs are loaded?

vborszeki: you've got a point, as the function name says, that the faqs based on context are loaded. however it also works without that, because the commands used after, wait for yielding the specific DOM element, but it seems more clear if we add the verification here. and of course it eliminates confusion. 😃

jaroslav-kubicek: great, with one note 

`cy.get('[data-cy=faq-categories]').should('exist');` - this should be first (to ensure some faqs are there, then check that "About Kiwi.com" doesn't exist)

vborszeki: Ah ok, then we were speaking about different things. 😄 I've added it here, to ensure that FAQs based on booking status are loaded, as the command name implies (which means that there are FAQs, but none of them is "About Kiwi.com"). 
What about, if we have both checks, for the inital load and for the ones depending on the context? 🤔

jaroslav-kubicek: my issue is all about to ensure that

1) some faqs are loaded (no loading or error is there)
2) none of the faqs is "about kiwi.com"

in order to ensure this, it should be done in this order, that's all

vborszeki: yepp, I understand. what I meant, is:

1) some faqs are loaded (no loading or error is there)
2) none of the faqs is "about kiwi.com"
3) some faqs are loaded (based on booking status, no loading or error is there)

what do you think about this?

vborszeki: @jaroslav-kubicek order fixed

PR #765, opened 2018-09-26
Do not show booking always

### LINKED ISSUE #733: "More info" link should not display another booking's details
As a user, I don't want to see another booking's info whole clicking on the "More info" link during the booking process.

Steps to reproduce the issue:
1. log in with any account that has at least one past or future booking
2. Search for a trip with a bus connection (e.g. Prague - Berlin)
3. Select on of the connections, go to booking page
4. Open the bus component and click "More info"

This is what you get:

snimek obrazovky 2018-09-14 v 7 58 07

Instead, we should hide the left panel (even thoug the user is logged in)  and display only the relevant info:

snimek obrazovky 2018-09-14 v 7 58 29

jaroslav-kubicek:
closes #733<br/><br/><br/><url>LiveURL: https://smartfaq-do-not-show-booking-always.surge.sh</url>


jaroslav-kubicek:
my job here is done, pls review

PR #769, opened 2018-09-27
feat(header): add translation keys

jalbertsr:
<br/><br/><br/><url>LiveURL: https://smartfaq-header-trasnaltion-keys.surge.sh</url>

### DIFF
--- a/src/common/layout/Header.js
+++ b/src/common/layout/Header.js
@@ -5,6 +5,7 @@ import * as React from 'react';
 import idx from 'idx';
 import css from 'styled-jsx/css';
 import { TextLink } from '@kiwicom/orbit-components';
+import Trans from '@kiwicom/nitro/lib/components/Text';
 
 import { Desktop, Mobile } from '../Responsive';
 import CloseButton from '../buttons/CloseButton';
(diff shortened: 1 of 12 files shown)


src/common/layout/Header.js
```diff
@@ -195,12 +206,14 @@ const renderLoggedOut = (
               onClick={() => push('/sign-in')}
               type="secondary"
             >
-              Sign&nbsp;In
```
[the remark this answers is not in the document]
vborszeki: fixed 🛠

src/common/layout/Header.js
```diff
@@ -190,7 +195,7 @@ const renderLoggedOut = (
         onClick={() => push('/sign-in')}
         type="secondary"
       >
-        Sign&nbsp;In
+        <Trans t={__('smartfaq.header.sign_in')} />
```
kantord: Shouldn't this be
```jsx
<Trans html t={__('smartfaq.header.sign_in')} />
```

Otherwise how would the non-breaking space be rendered?

vborszeki: I thought it was already added in my last commit, don't know what happened here 🙀

PR #770, opened 2018-09-27
feat(select-another-booking): add translation keys

kantord:
<br/><br/><br/><url>LiveURL: https://smartfaq-select-another-booking-translation-key.surge.sh</url>


src/common/booking/SelectAnotherBookingLink.js
```diff
@@ -39,7 +40,7 @@ const SelectAnotherBookingLink = (props: Props) => (
                 external={false}
                 type="primary"
               >
-                Select another booking
+                <Trans t={__('smartfaq.select_another_booking')} />
```
vborszeki: I see, that there is no more translated text in this file, but maybe for consistency I would use something like `smartfaq.select_another_booking.link`. What do you think?

kantord: I don't know. Does it make a difference? :D

PR #778, opened 2018-09-27
feat(common): translation keys for common components

jalbertsr:
<br/><br/><br/><url>LiveURL: https://smartfaq-common-components-keys-trans.surge.sh</url>


src/common/ContactPageLink.js
```diff
@@ -19,7 +20,7 @@ const ContactPageLink = ({
   return (
     <div className="contactUs">
       <a target="_blank" href={`/${language}/content/feedback`}>
-        {text}
+        <Trans t={text} />
```
vborszeki: I would rename this _text_ prop to something like _translationKey_ and then I would pass consistently just translation keys to `<ContactPageLink />` component.
There are some occurrences now, where just simple strings are passed (like in src/SingleBookingPage/bookingItem/Contact.js), which wont't be translated, and the default value for the _text_ prop itself is just a static string too.
What do you think?

kantord: I agree with @vborszeki

jalbertsr: Agree, thanks for the suggestion.

jalbertsr: fixed.

